PR #18, opened 2020-05-16
Implemented GithubRepo

### LINKED ISSUE #15: GithubSelf.java:38-40: Implement User.repos(). It should...
The puzzle `13-b311c7f8` from #13 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/e8313b595e44eb3d86868787cc87a41613adf5a3/self-core-impl/src/main/java/com/selfxdsd/core/GithubSelf.java#L38-L40

The puzzle was created by @amihaiemil on 16-May-20. 

amihaiemil:
Resolves #15 
Implemented and tested GithubRepo.

Left puzzles for continuing with Github Repos imlementation, as well as GithubRepo.activate().

### DIFF
--- a/self-api/pom.xml
+++ b/self-api/pom.xml
@@ -13,5 +13,12 @@
 
     <artifactId>self-api</artifactId>
 
+    <dependencies>
+        <dependency>
+            <groupId>javax.json</groupId>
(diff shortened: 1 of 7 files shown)


amihaiemil:
@rultor merge it

Issue #20, opened 2020-05-16
GithubSelf.java:37-40: Implement User.repos(). It should...


0pdd:
The puzzle `15-5c5b51d8` from #15 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/62a0f94ea7fe212f4f331a49fe49760a7d32038f/self-core-impl/src/main/java/com/selfxdsd/core/GithubSelf.java#L37-L40

The puzzle was created by @amihaiemil on 16-May-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

0pdd:
The puzzle `15-5c5b51d8` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #25, opened 2020-05-17
#20 Repos replaced with Provider

amihaiemil:
PR for #20 

* Interface ``Repos`` replaced with ``Provider``. At the moment we are interested only in the public repositories which can be fetched and displayed on the front-end, while the backend expects only the repo's name to fetch it from the provider;

* ``Github`` is the first Provider we have;
* Left puzzle for continuing with method ``Github.repo()``;


amihaiemil:
@rultor merge it

PR #34, opened 2020-05-18
#33 cache the repo json

### LINKED ISSUE #33: Fetch the json on Repo Creation
At the moment the json is fetched when calling ``Repo.json()``. It's better if we fetch it on creation, because the ``json()`` method might be called multiple times.

amihaiemil:
Fixes #33 

Decided to cache the result of this.json(), instead of fetching the value in the constructor.


amihaiemil:
@rultor merge

PR #35, opened 2020-05-18
#21 BaseRepo.activate() implemented and tested

amihaiemil:
PR for #21 
Wrote only a shallow unit test which should be revised once possible.


amihaiemil:
@rultor merge it

PR #43, opened 2020-05-19
#41 storage package for the API

### LINKED ISSUE #41: Create package storage in slef-api
Since the Storage is quite a separate concern and it also has a few concrete classes, it should be in its own package: ``com.selfxdsd.api.storage``

amihaiemil:
fixes #41


amihaiemil:
@rultor merge

Issue #36, opened 2020-05-18
Start implementing GitLab Provider.


amihaiemil:
Following the pattern of the Github Provider, we should implement one for GitLab as well. First, we should check the API of GitLab and see how we can fetch a Repo, assign Issues etc.

criske:
I'll take it

amihaiemil:
@criske I suggest to start with very little code. First, check GitLab's RESTful API and see how we can fetch a Repo, its Issues etc. 

Document what you find here, in this ticket, please :)

criske:
To get all repos owned by an autheticated user:  https://gitlab.com/api/v4/projects/?owned=1
To get one repo:   https://gitlab.com/api/v4/projects/:repoId

**Very important**: all these calls must have in their header either a private-token or a oauth access- token. I've tested with private-token
https://docs.gitlab.com/ee/api/#personalproject-access-tokens

amihaiemil:
@criske and is there a way to get the **public** repos, or one public repo of a User without authentication? GitHub allows us to do that, I think GitLab should allow it as well :D

criske:
Is possible, but in that case I think we'll have to use the gitlab domain associated to that user (like: gitlab.example.com), and that will brake the User API. 
I will dig more, I'm new to Gitlab.

amihaiemil:
@criske 

>  but in that case I think we'll have to use the gitlab domain associated to that user (like: gitlab.example.com)

No, I don't think so. We'll work only with the public GitLab, at GitLab.com.

amihaiemil:
@criske As far as I can see, we can read the public Repos with by accessing https://gitlab.com/api/v4/projects/ 

So surely, there must be a way to get the public repos of a user, too. 

It's enough info for now, thanks! I'll send you the bounty for doing this documentation work and then we can start with the code.

criske:
Without auth, only public repos:  https://gitlab.com/api/v4/users/:userName/projects/

amihaiemil:
@criske perfect, thanks

criske:
One more thing, just checked: if we attach the access token to that endpoint, it will list the private repos too.

amihaiemil:
@criske Thank you for solving this!

PR #46, opened 2020-05-20
Implemented Gitlab and GitlabRepo.

criske:
PR for #36 
Integration tests for Gitlab over network. 
Tested finding a repo and its json format.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/Gitlab.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/Gitlab.java
@@ -1,0 +1,88 @@
+/**
+ * Copyright (c) 2020, Self XDSD Contributors
+ * All rights reserved.
+ * <p>
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"),
+ * to read the Software only. Permission is hereby NOT GRANTED to use, copy,
+ * modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software.
+ * <p>
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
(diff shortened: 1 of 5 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/Gitlab.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.net.URI;
+
+/**
+ * Gitlab as a Provider.
+ * @author criske
+ * @version $Id$
+ * @since 0.0.1
+ *  so we can mock the Gitlab server and write unit tests for the methods
+ *  that perform HTTP Requests.
```
amihaiemil, line 38: @criske Can you remove these 2 lines and leave a puzzle for Integration Testing?

self-core-impl/src/test/java/com/selfxdsd/core/GitlabTestCase.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.json.JsonObject;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Integration tests for {@link Gitlab}.
```
amihaiemil, line 15: @criske Well, I would say this class is a class of unit tests, not integration tests :) 
Conceptually, yes, these are integration tests because there are more than 1 real class involved.

But in the case of this project, we call Integration Tests, the tests that actually call the Gitlab API.
For instance, see how GithubITCase is implemented.

So, can you please correct the javadoc and say "unit tests for ..."?

self-core-impl/src/test/java/com/selfxdsd/core/GitlabTestCase.java
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @param username Provided user name
+     * @return Created Gitlab instance
+     */
+    private Gitlab createGitlab(final String username) {
+        final User user = mock(User.class);
+        when(user.username()).thenReturn(username);
+        when(user.provider()).then(invocation -> new Gitlab(
+                (User) invocation.getMock(),
+                null));
```
amihaiemil, line 55: @criske Here and in the files above, can you please correct the indentation? We use 4-space indentation, and here are 8 spaces. Thanks.

self-core-impl/src/test/java/com/selfxdsd/core/GitlabTestCase.java
criske: Actually those tests are using the real Gitlab API call, since BaseRepo#json() is tight coupled with with the real network layer. I've used dummy projects on my gitlab account to test.

amihaiemil: @criske oh, I didn't notice. That's fine as well, but please rename the test file from ``*TestCase`` to ``*ITCase`` -- ``ITCase`` is the suffix used by Maven to run integration tests. 

Integration tests are only run when we specify the ``-Pitcases`` to Maven.

self-core-impl/src/test/java/com/selfxdsd/core/GitlabTestCase.java
criske: Sure, I'll fix all when I get to my machine. In about 30 min :)

amihaiemil:
@criske great, thanks! I think the puzzle is a bit too short, maybe 0pdd will complain. But let's see.

amihaiemil:
@rultor merge it please

PR #58, opened 2020-05-21
Implemented InMemoryContracts

criske:
-updated Contributor core API to have id;
-tested InMemoryContracts;
-tested Project#contracts() and Contributor#contracts() via InMemoryContracts;

### DIFF
--- a/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
+++ b/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
@@ -22,23 +22,32 @@
  */
 package com.selfxdsd.core.mock;
 
-import com.selfxdsd.api.Contract;
-import com.selfxdsd.api.Contracts;
+import com.selfxdsd.api.*;
 import com.selfxdsd.api.storage.Storage;
 
-import java.util.Iterator;
+import java.math.BigDecimal;
+import java.util.*;
 
 /**
  * In-Memory contracts for testing purposes.
+ *
  * @author Mihai Andronache (amihaiemil@gmail.com)
  * @version $Id$
  * @since 0.0.1
- * @todo #42:30min Implement this class and write some unit tests
- *  for Project.contracts(). Then go further and implement
- *  Contributor.contracts()
  */
 public final class InMemoryContracts implements Contracts {
-
+    /**
+     * In memory holder of contracts.
+     */
+    private final Set<Contract> contracts = new HashSet<>();
+    /**
+     * In memory holder of projects.
+     */
+    private final Set<ProjectInternal> projects = new HashSet<>();
+    /**
+     * In memory holder of contributors.
+     */
+    private final Set<ContributorInternal> contributors = new HashSet<>();
     /**
      * Parent storage.
      */
@@ -46,7 +55,8 @@ public final class InMemoryContracts implements Contracts {
 
     /**
      * Constructor.
-     * @param storage Parent storage.
+     *
+     * @param storage Parent storage
      */
     public InMemoryContracts(final Storage storage) {
         this.storage = storage;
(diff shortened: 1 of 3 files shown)


self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,378 @@
...
+     */
+    private final Set<Contract> contracts = new HashSet<>();
+    /**
+     * In memory holder of projects.
+     */
+    private final Set<ProjectInternal> projects = new HashSet<>();
+    /**
+     * In memory holder of contributors.
+     */
+    private final Set<ContributorInternal> contributors = new HashSet<>();
```
amihaiemil, line 50: @criske But the Project and the Contrinbutor are supposed to be found on the Contract. Contract already has methods ``project()`` and ``contributor()``. Why do we need a Set of projects and a Set of Contributor?

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,378 @@
...
+
+    /**
+     * In memory implementation of a {@link Project}.
+     * <br>
+     * Aids {@link InMemoryContracts}
+     * <br>
+     * Has capability to link a contract.
+     * See: {@link ProjectInternal#addContract(Contract)}.
+     */
+    private static final class ProjectInternal implements Project {
```
amihaiemil, line 150: @criske I think ProjectInternal and ContributorInternal should be similar to the ``StoredProject`` and ``StoredProjectManager`` we have.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,378 @@
...
+         * Global contracts. Access to {@link InMemoryContracts }
+         * Grants access to any arbitrary project's contracts using
+         * {@link Contracts#ofProject(int) from any point of the object graph.
+         */
+        private final Contracts rootContracts;
+
+        /**
+         * Contracts associated to this project.
+         */
+        private final Set<Contract> ownedContracts = new HashSet<>();
```
amihaiemil, line 166: @criske We don't really need this Set. The Project already has the ``rootContracts``, which should have an ``add`` method... but we can leave it as it is for now, until we implement ``Contracts.register``, or something.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,378 @@
...
+        }
+
+        @Override
+        public Repo repo() {
+            return null;
+        }
+
+        @Override
+        public Contracts contracts() {
+            return new Contracts() {
```
amihaiemil, line 206: @criske I think ``return this.rootContracts.ofProject(projectId)`` is enough here.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,378 @@
...
+        public int contributorId() {
+            return id;
+        }
+
+        @Override
+        public Contracts contracts() {
+            return new Contracts() {
+                @Override
+                public Contracts ofProject(final int projectId) {
+                    return rootContracts.ofProject(projectId);
```
amihaiemil, line 287: @criske this is wrong here, it should be ``rootContracts.ofContributor(...)``, but we don't have this method yet

criske:
I think you should reject this PR @amihaiemil .Tthe implementation of InMemoryContracts is done in isolation of InMemoryProjects. Need to integrate with that.

amihaiemil:
@criske 
> .Tthe implementation of InMemoryContracts is done in isolation of InMemoryProjects. Need to integrate with that.

Yes, I think that's you use those sets and internal classes? Make a smaller PR, small steps, so we're sure the architecture is followed :D

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
criske: we will get stack-overflow

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
criske: I was following the Contracts API :)

criske:
Anyway sorry for this big PR, but I can't think of a way to make it smaller just using the current API :)

PR #61, opened 2020-05-21
Issue53

criske:
Integrated InMemoryContracts with Storage

### DIFF
--- a/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
+++ b/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
@@ -22,23 +22,32 @@
  */
 package com.selfxdsd.core.mock;
 
-import com.selfxdsd.api.Contract;
-import com.selfxdsd.api.Contracts;
+import com.selfxdsd.api.*;
 import com.selfxdsd.api.storage.Storage;
 
-import java.util.Iterator;
+import java.math.BigDecimal;
+import java.util.*;
 
 /**
  * In-Memory contracts for testing purposes.
+ *
  * @author Mihai Andronache (amihaiemil@gmail.com)
  * @version $Id$
  * @since 0.0.1
- * @todo #42:30min Implement this class and write some unit tests
- *  for Project.contracts(). Then go further and implement
- *  Contributor.contracts()
  */
 public final class InMemoryContracts implements Contracts {
-
+    /**
+     * In memory holder of contracts.
+     */
+    private final Set<Contract> contracts = new HashSet<>();
+    /**
+     * In memory holder of projects.
+     */
+    private final Set<ProjectInternal> projects = new HashSet<>();
+    /**
+     * In memory holder of contributors.
+     */
+    private final Set<ContributorInternal> contributors = new HashSet<>();
     /**
      * Parent storage.
      */
@@ -46,7 +55,8 @@ public final class InMemoryContracts implements Contracts {
 
     /**
      * Constructor.
-     * @param storage Parent storage.
+     *
+     * @param storage Parent storage
      */
     public InMemoryContracts(final Storage storage) {
         this.storage = storage;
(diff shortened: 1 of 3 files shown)


self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,392 @@
...
+     */
+    private final Set<Contract> contracts = new HashSet<>();
+    /**
+     * In memory holder of projects.
+     */
+    private final Set<ProjectInternal> projects = new HashSet<>();
+    /**
+     * In memory holder of contributors.
+     */
+    private final Set<ContributorInternal> contributors = new HashSet<>();
```
amihaiemil, line 51: @criske Why do we need these 2 sets?
The Project and the Contributor should be taken from the Contract.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,392 @@
...
+    /**
+     * In memory implementation of a {@link Project}.
+     * Wraps a real implementation of a Project.
+     * <br>
+     * Aids {@link InMemoryContracts}
+     * <br>
+     * Has capability to link a contract.
+     * See: {@link ProjectInternal#addContract(Contract)}.
+     */
+    private static final class ProjectInternal implements Project {
```
amihaiemil, line 163: @criske I think these Internal classes are dummy classes to be used until we have some concrete implementations of Project and Contributor, right? :D

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,392 @@
...
+        }
+
+        @Override
+        public Repo repo() {
+            return delegate.repo();
+        }
+
+        @Override
+        public Contracts contracts() {
+            return new Contracts() {
```
amihaiemil, line 220: @criske I think you can simply return ``rootContracts.ofProject(...)``

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,392 @@
...
+        }
+
+        @Override
+        public int contributorId() {
+            return id;
+        }
+
+        @Override
+        public Contracts contracts() {
+            return new Contracts() {
```
amihaiemil, line 298: @criske THis method should throw UnsupportedOperationException at the moment, because Contracts.ofContributor(...) does not exist yet.

amihaiemil:
@criske I also noticed you never used the encapsulated ``Storage``. Why not?

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
criske: We need to link a Contract to a Project and Contributor and vice -versa. Those sets are "linked" projects and contributors with their contracts. 
In Contract when we get the Project and then searching in Project#contracts() we must find the same Contract.
Same for Contributor.
See my unit tests.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
criske: We need the "owned" contracts when iterate this project contracts, that's why I've used this proxy. 
rootContracts.ofProject is "global" search for any valid project id

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
criske: but then it will fail Contributor#contracts() tests

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
```diff
@@ -22,43 +22,392 @@
...
+     * @param projectId Project's
+     * @return Internal representation of a Project
+     */
+    private ProjectInternal findOrCreateProjectInternal(final int projectId) {
+        ProjectInternal projectInternal = projects.stream()
+            .filter(p -> p.projectId() == projectId)
+            .findFirst()
+            .orElse(null);
+        if (projectInternal == null) {
+            Project projectFromStorage = StreamSupport
```
criske, line 117: Storage is used here

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java
criske: Also if we  just call rootContracts.ofProject(this.projectId()) instead, we will get stack overflow

amihaiemil:
@criske 

> Anyway sorry for this big PR, but I can't think of a way to make it smaller just using the current API :)

You can always split a task into multiple smaller tasks. It's ok to implement a class, a method, and make the rest of the method throw ``UOE`` -- then, add a ``todo`` to continue work in a future task.

I'll merge this, even though it's not what I had mind. But it's not your fault, the task was quite broad, with little details. It's a mock class anyway, we can change it later.

amihaiemil:
@rultor merge it please

amihaiemil:
@rultor try to merge again

criske:
When you have time, could share how would you tackle the problem, lets say for contract-project() -contracts() in the InMemoryContracts scenario. I might learn a thing or two for the future :D

amihaiemil:
@criske sure, no problem. Right now I'm banging my head with another thing :D

Issue #96, opened 2020-05-24
InMemoryContracts.java:41-43: Implement and test Project...


0pdd:
The puzzle `79-2e30e1fe` from #79 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/7380b67b3c6869df69708cd5b38d83a860575992/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContracts.java#L41-L43

The puzzle was created by @amihaiemil on 24-May-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

BogdanAdrian11:
I'll take this.

amihaiemil:
@BogdanAdrian11 paid with transaction id 8527787064150430H

0pdd:
The puzzle `79-2e30e1fe` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #103, opened 2020-05-24
#96 - Implement Projects getByProjectId method.

BogdanAdrian11:
Pull request for #96 .
Implemented getByProjectId method and unit test.
Left puzzle in InMemoryProjectsTestCase to continue tests for other types of projects.

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Projects.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Projects.java
@@ -22,6 +22,8 @@
  */
 package com.selfxdsd.api;
 
+import java.util.Iterator;
+
 /**
  * Projects managed by Self.
  * @author Mihai Andronache (amihaiemil@gmail.com)
@@ -51,4 +53,22 @@ public interface Projects extends Iterable<Project> {
      * @return Projects.
      */
     Projects ownedBy(final User user);
+
+    /**
+     * Get the Project with the corresponding projectId.
+     * @param projectId Id of the Project.
+     * @return Project, or null if no Project with this id is found.
+     */
+    default Project getProjectById(final int projectId) {
+        Iterator<Project> iterator = iterator();
+        Project project = null;
+        while (iterator.hasNext()) {
+            Project nextProject = iterator.next();
+            if (nextProject.projectId() == projectId) {
+                project = nextProject;
+                break;
+            }
+        }
+        return project;
+    }
 }
--- a/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryProjectsTestCase.java
+++ b/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryProjectsTestCase.java
@@ -20,8 +20,9 @@ import static org.mockito.Mockito.when;
  * @author hpetrila
  * @version $Id$
  * @since 0.0.1
- * @todo #96:30min Continue tests for {@link Projects#getProjectById(int)}
- *  in PmProjectsTestCase and UserProjectsTestCase.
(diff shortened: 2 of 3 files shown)


self-api/src/main/java/com/selfxdsd/api/Projects.java
```diff
@@ -51,4 +53,22 @@
      * @return Projects.
      */
     Projects ownedBy(final User user);
+
+    /**
+     * Get the Project with the corresponding projectId.
+     * @param projectId Id of the Project.
+     * @return Project, or null if no Project with this id is found.
+     */
+    default Project getProjectById(final int projectId) {
```
amihaiemil, line 62: @BogdanAdrian11 This method cannot be default, because the base implementation of ``Projects`` will represent *all* projects in the system, and we will not be able to iterate over all of them (loading them all into memory will probably be not possible). 

But it's ok to leave it like this for now. Just specify in the ``TODO`` that this method should be implemented by all implementors of Projects.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryProjectsTestCase.java
```diff
@@ -81,6 +83,37 @@ public void projectOwnedBy() {
...
+     * Check if no project is returned when id is not found.
+     */
+    @Test
+    public void getProjectByNotFoundId() {
+        final Storage storage = new InMemory();
+        ProjectManager projectManager = storage
+                .projectManagers().pick("github");
+        InMemoryProjects projects = (InMemoryProjects)
+                storage.projects();
+        projects.register(mock(Repo.class), projectManager);
```
amihaiemil, line 112: @BogdanAdrian11 you don't need to register any project for this test. Just a ``getById`` with ``-1`` is logically enough.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryProjectsTestCase.java
```diff
@@ -81,6 +83,37 @@ public void projectOwnedBy() {
...
 
+    /**
+     * Check if the right project is returned.
+     */
+    @Test
+    public void getProjectById() {
+        final Storage storage = new InMemory();
+        ProjectManager projectManager = storage
+                .projectManagers().pick("github");
+        InMemoryProjects projects = (InMemoryProjects)
```
amihaiemil, line 94: @BogdanAdrian11 you don't need any cast. ``Projects all = storage.projects()`` is enough.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryProjectsTestCase.java
```diff
@@ -20,6 +20,8 @@
  * @author hpetrila
  * @version $Id$
  * @since 0.0.1
+ * @todo #96:30min Continue tests for {@link Projects#getProjectById(int)}
+ *  in PmProjectsTestCase and UserProjectsTestCase.
```
amihaiemil, line 24: @BogdanAdrian11 Here, please update this puzzle to specify that each implementation of ``Projects`` should provide their own ``getById`` method. 

For instance, in the case of ``UserProjects``, the method will be allowed to use ``iterator()`` since we can afford to load all of a User's projects in memory and iterate them.

amihaiemil:
@BogdanAdrian11 Looks nice thanks! Sorry for the delayed answer, I noticed only now that you made the changes already.

amihaiemil:
@rultor merge it

Issue #95, opened 2020-05-24
InMemoryContributors.java:40-42: Write some unit tests...


0pdd:
The puzzle `79-282aa471` from #79 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/7380b67b3c6869df69708cd5b38d83a860575992/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContributors.java#L40-L42

The puzzle was created by @amihaiemil on 24-May-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

HoreaPetrila:
I'll do it

0pdd:
The puzzle `79-282aa471` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

amihaiemil:
@HoreaPetrila paid with transaction id 6RP99907P46129158

PR #105, opened 2020-05-25
#95 - Implement InMemoryContributorsTestCase

HoreaPetrila:
PR For #95

### DIFF
--- a/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContributorsTestCase.java
+++ b/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContributorsTestCase.java
@@ -45,6 +45,17 @@ public final class InMemoryContributorsTestCase {
     }
 
     /**
+     * Get contributor by an unknown id should return null.
+     */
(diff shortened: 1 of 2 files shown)


self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryContributorsTestCase.java
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    public void contributorGetById() {
+        final Storage storage = new InMemory();
+        Contributor registered = storage.contributors()
+            .register("horea", "github");
+        assertThat(storage.contributors()
+            .getById("horea", "github"),
+            is(registered));
+
+    }
```
amihaiemil, line 45: @HoreaPetrila Can you also add a unit test to check that ``getById`` returns ``null`` when the Contributor does not exist?

amihaiemil:
@HoreaPetrila looks good, thanks!

amihaiemil:
@rultor merge it

Issue #101, opened 2020-05-24
GithubIssues.java:37-40: Implement and unit test method...


0pdd:
The puzzle `97-8b9168b0` from #97 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/fdbd1d044d9248d08980205b75dd71cc9e86f974/self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java#L37-L40

The puzzle was created by @amihaiemil on 24-May-20. 

 role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
Sign me up

amihaiemil:
paid with transaction id: 6EV87513ED468354G

0pdd:
The puzzle `97-8b9168b0` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #106, opened 2020-05-25
Implemented and tested method json() in GithubIssues

criske:
PR for #101

Added an internal utility class to fetch json objects remotely.
Added a temporary cache for fetched issues until Storage has that feature.

Created puzzle to extract the json utility class into its own file and then refactor BaseRepo#json to use it.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
@@ -117,7 +117,7 @@ public final class GithubIssues implements Issues {
          * Private constructor.
          */
         private JsonFetcher() {
-            throw new IllegalStateException("Singleton");
+            throw new UnsupportedOperationException();
         }
 
         /**
--- a/self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
@@ -35,17 +35,13 @@ import java.net.URI;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
-import java.util.HashMap;
 import java.util.Iterator;
-import java.util.Map;
 
 /**
  * Issues in a Github repository.
  *
  * @author Mihai Andronache (amihaiemil@gmail.com)
  * @version $Id$
- * @todo #98 Extract the utility class `GithubIssues.JsonFetcher` into its
- *  own file and then refactor `BaseRepo#json()` to use it.
  * @since 0.0.1
  */
 public final class GithubIssues implements Issues {
@@ -61,11 +57,6 @@ public final class GithubIssues implements Issues {
     private Storage storage;
 
     /**
-     * Temporary cache until {@link Storage} has that feature.
-     */
-    private final Map<Integer, Issue> cachedIssues = new HashMap<>();
-
-    /**
      * Ctor.
      *
      * @param issuesUri Issues base URI.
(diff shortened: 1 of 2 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
```diff
@@ -72,4 +104,52 @@ public Issue getById(final int issueId) {
...
         );
     }
+
+    /**
+     * Utility class used for fetching json objects over the network.
+     * @author criske
+     * @version $Id$
+     * @since 0.0.1
+     */
+    private static final class JsonFetcher {
```
amihaiemil, line 114: @criske No utility classes, please. Just implement the call into the json() similarly to how BaseRepo json works.

self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
```diff
@@ -63,7 +78,24 @@ public GithubIssues(final URI issuesUri, final Storage storage) {
 
     @Override
     public Issue getById(final int issueId) {
-        throw new UnsupportedOperationException("Not yet implemented.");
+        Issue issue = cachedIssues.get(issueId);
+        if (issue == null) {
```
amihaiemil, line 82: @criske No need to cache the Issue. It is also not correct to cache it since this Issue is live on Github/GitLab etc and could have it's state (e.g. assignee) changed at any time.

Just fetch the Json and return a ``GithubIssue`` or null if the HTTP call returns 404 NOT FOUND or 204 NO CONTENT.

self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
criske: Well I know that, but I thought to use an utility class until we have a network abstraction, we discussed that before. Cause right know we just copy/paste the same code for json objects fetching -  basically introducing the same tight-coupling as an utility class  + DRY violation.
But sure I'll remove that :)

self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
criske: Ok

self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
amihaiemil: Yes, maybe we'll add some networking abstraction later on. But right now, it's better to have a bit of duplicated code (it's code in different objects after all), than having utility classes.

Once someone sees a utility class, the whole concept of encapsulation goes to hell. 

I believe Encapsulation is the most important thing, above all else...

self-core-impl/src/main/java/com/selfxdsd/core/issues/GithubIssues.java
criske: One question, what should we return when we get other codes (like 500s) or an IOException? We throw?

amihaiemil: @criske wrap it inside ISE with a meaningful message and throw it. Similarly to this: https://github.com/self-xdsd/self-core/blob/master/self-core-impl/src/main/java/com/selfxdsd/core/BaseRepo.java#L109

criske: > @criske No need to cache the Issue. It is also not correct to cache it because this Issue is live on Github and could have its state (e.g. assignee) changed at any time.
>
Just got an idea about caching, for the future: we could with etags if an issue has changed. I'm pretty sure that github supports etags.

amihaiemil: @criske maybe, but I'm not sure it would help a lot. We'll see :D

amihaiemil:
@rultor merge it please

PR #109, opened 2020-05-25
Role is part of the Contract PK

### LINKED ISSUE #107: InMemoryContracts: the role should be part of the key
A Contributor can have multiple contracts with the same Project: 1 contract for DEV work, 1 for ARCH work etc.

This means, the role from a Contract should be part of the composite primary key, together with the projectId and the contributorId (username + provider).

amihaiemil:
fixes #107 

* role is part of the Contract PK
* unit tests
* added constants for the possible roles


amihaiemil:
@rultor merge it please

Issue #136, opened 2020-05-29
Tasks.java:30-31: Implement and unit test...


0pdd:
The puzzle `133-7376f99a` from #133 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/d8f98bcb2e001d80c36155956d22cd836c246efc/self-api/src/main/java/com/selfxdsd/api/Tasks.java#L30-L31

The puzzle was created by @amihaiemil on 29-May-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

amihaiemil:
waiting for #134

criske:
No worries about the bounty :D. I was about to take one of the issues
tomorrow morning. But sure assign me, I'll take care of it this evening, if
not, tomorrow morning for sure.

amihaiemil:
@criske whenever you want/have time, it's not urgent.

0pdd:
The puzzle `133-7376f99a` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #139, opened 2020-05-29
#136 Implemented and unit tested Contributor.tasks()

criske:
PR for #136 

Added `ofContributor(UserName,ProviderName)` in Tasks.
Stubbed `InMemoryTasks#ofContributor()`
Contributor#tasks() impl will return ContributorTasks via `storage.tasks.ofContributor`.
Implemented and tested `ContributorTasks`.
Implemented and tested `StoredContributor#tasks()`.

Removed puzzle.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContributorTasks.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContributorTasks.java
@@ -68,15 +68,7 @@ public final class ContributorTasks implements Tasks {
 
     @Override
     public Task register(final Issue issue) {
-        if (!this.provider.equals(issue.provider())) {
-            throw new IllegalArgumentException(
-                "The given Issue does not belong to the Provider "
-                    + this.provider + "."
-            );
-        }
-        Task task = this.storage.tasks().register(issue);
-        tasks.add(task);
-        return task;
+        throw new UnsupportedOperationException("Not implemented yet");
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContributorTasks.java
```diff
@@ -0,0 +1,111 @@
...
+    public Task register(final Issue issue) {
+        if (!this.provider.equals(issue.provider())) {
+            throw new IllegalArgumentException(
+                "The given Issue does not belong to the Provider "
+                    + this.provider + "."
+            );
+        }
+        Task task = this.storage.tasks().register(issue);
+        tasks.add(task);
+        return task;
```
amihaiemil, line 79: @criske I think you should leave this method unimplemented for now (throw UOE) because it should be a little more complex: first, check if the Issue and Contributor belong to the same Project, then add the task in Storage and also **Assign** it to the Contributor (this part is not yet implemented at all).

And it gets a little more complicated because the Task in the DB is actually linked to the Contract (based on role), not to the Contributor. So the link between Task and Contributor is indirect.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryTasks.java
```diff
@@ -110,6 +107,11 @@ public Tasks ofProject(
         return null;
     }
 
+    @Override
+    public Tasks ofContributor(final String username, final String provider) {
+        throw new UnsupportedOperationException("Not implemented yet");
```
amihaiemil, line 112: @criske Can you leave a puzzle so this method will be implemented later?

amihaiemil:
@criske thanks!

amihaiemil:
@rultor merge it

PR #145, opened 2020-05-30
StoredUser implemented

### LINKED ISSUE #142: Implement StoredUser
Right now we only have User as anonymous classes. Let's implement StoredUser.

amihaiemil:
Fixes #142 

* removed User.avatar(). No need to store the avatar URL in the DB.
* implemented and tested StoredUser
* Used it instead of anonymous classes


amihaiemil:
@rultor merge

PR #156, opened 2020-06-01
#155 Users.signUp(...) refactored

### LINKED ISSUE #155: Users.singUp(...) should take params instead of User
It doesn't make sense to expect the User as input to the method ``signUp``. Instead, this method should expect the parameters from which it makes up a User and returns it.

amihaiemil:
fixes #155


amihaiemil:
@rultor merge it please

PR #157, opened 2020-06-02
#152 StoredProject constructor redesigned

### LINKED ISSUE #152: StoredProject should take the User and String repoFullName
At the moment, when building a StoredProject, you have to have the ``Repo``. However, it would be more efficient to pass the User ``owner`` and the ``repoFullName`` instead, and call Github's API on demand only if ``project.repo()`` is called.

amihaiemil:
fixes #152


amihaiemil:
@rultor merge it

Issue #153, opened 2020-05-31
Make `accessToken` available to User API


criske:
Right now `accessToken` is used internally by StoredUser, but `self-storage` needs the token during 
Users#signUp(User) (to insert/update into db).

amihaiemil:
@criske you're right. I'll fix it tomorrow together with another ticket and deploy version ``0.0.2-SNAPSHOT`` -- from this version on we're also deploying javadocs and sources so development of the storage becomes easier (your IDE will display proper parameter names)

amihaiemil:
@criske On second thought, I don't think we should save this token in the DB. 

Because not everyone who gets hold of a User should have access to their Github protected resources (for instance, a PM will haal ve access to a Project's owner which is a User, but they should not be able to use the Github provider on behalf of the User). 

The solution should be a decorator over ``User``, called ``Authenticated``, which injects the token into the Provider when ``User.provider()`` is called.

So ``SelfUsers.signUp`` will look like this:

```java
public User signUp(...){
    final User user = ...;
    return new Authenticated(user, access_token);
}
```

And then:

```java
public final class Authenticated implements User {
    ...
    public Provider provider() {
        return this.provider.authenticated(this.access_token);
    }
}
```

I'll start implementing it.

amihaiemil:
@criske see PR #158: the user's token is not saved in the DB and it is also not exposed to anyone. 
Only the ``SelfCore.login`` method will provide an Authenticated user if the token is present, which in turn will pass the token to the Provider.

This mechanism makes sure that nobody else from the system has access to a User's protected resources.

PR #158, opened 2020-06-02
Implemented Authenticated user

amihaiemil:
PR for #153, based on [this](https://github.com/self-xdsd/self-core/issues/153#issuecomment-636643821) comment.

* Implemented Authenticated decorator for a user who has an accessToken;
* Implemented Provider.withToken(...) to return a new version of it which contains the token.


amihaiemil:
@rultor merge it please

Issue #161, opened 2020-06-04
Implement the Invoice


amihaiemil:
A Contract should have a corresponding Invoice containing all the tasks that have been finished by the Contributor.

criske:
Sign me up

amihaiemil:
@criske Actually, it's a 1-to-many relationship. A Contract has more Invoices, but only one of them is "active" (not yet paid) -- we will always add a finished task to the active Invoice, the one that is not yet paid. 

If all the invoices are paid, we create a new one.

amihaiemil:
payment done with transaction id: 46151150Y9859444R

PR #163, opened 2020-06-06
#161 Invoice/Invoices API for Contract.

criske:
PR for #161 

Added puzzle to implement and test InMemoryInvoices; also test the Invoice/Invoices API.

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Contract.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Contract.java
@@ -23,6 +23,7 @@
 package com.selfxdsd.api;
 
 import java.math.BigDecimal;
+import java.util.Objects;
 
 /**
  * A collaboration Contract between a Project and a Contributor.
@@ -55,6 +56,14 @@ public interface Contract {
      * @return String.
      */
     String role();
+
+    /**
+     * Invoices for this contract, active or inactive.
+     * <br>
+     * Note that a contract must have at most one active Invoice.
+     * @return Iterable of Invoice.
+     */
+    Invoices invoices();
 
     /**
      * Possible roles in a Contract.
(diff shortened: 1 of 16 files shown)


self-api/src/main/java/com/selfxdsd/api/Contract.java
```diff
@@ -101,4 +110,74 @@ private Roles(){ }
...
+         */
+        public final String contributorUsername;
+        /**
+         * Contributor/Project's provider.
+         */
+        public final String provider;
+        /**
+         * Contributor's role.
+         */
+        public final String role;
```
amihaiemil, line 135: @criske Please make these private. We should never have public attributes, no matter what.

self-core-impl/src/main/java/com/selfxdsd/core/contracts/invoices/StoredInvoice.java
```diff
@@ -0,0 +1,84 @@
...
+    public BigDecimal totalAmount() {
+        final Contract contract = this.storage.contracts().findById(contractId);
+        BigDecimal totalAmount = BigDecimal.ZERO;
+        if(contract != null){
+            final BigDecimal rate = contract.hourlyRate();
+            totalAmount = this.tasks().stream()
+                .reduce(BigDecimal.ZERO,
+                    (acc, curr)-> acc.add(rate.multiply(BigDecimal
+                        .valueOf(curr.timeSpent().toHours()))),
+                    (acc, curr)-> acc);
```
amihaiemil, line 74: @criske Can you please not use the Stream API here? This method should return a sum, but it looks very strange right now... what does the ``reduce`` method do?

self-api/src/main/java/com/selfxdsd/api/Invoices.java
```diff
@@ -0,0 +1,64 @@
...
+     * @return Created InvoiceTask
+     */
+    InvoiceTask add(Task task, Duration timeSpent);
+
+    /**
+     * Checks if invoice is active.
+     * @param id Invoice id.
+     * @return Boolean.
+     */
+    boolean isActive(int id);
```
amihaiemil, line 46: @criske Can you please rename this mathod to ``isPaid``?

self-api/src/main/java/com/selfxdsd/api/Invoices.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Proceed to pay the contract's contributor.
+     * After payment confirmed, invoice will became inactive.
+     * @param id Invoice id.
+     * @return Inactive Invoice.
+     */
+    Invoice deactivate(int id);
```
amihaiemil, line 54: @criske I would remove this method for now. In principle it is ok, but it's too early for it, I have no idea how the payment mechanism will work yet :D

self-api/src/main/java/com/selfxdsd/api/Contract.java
criske: Why? As long the fields are final primitives/immutable is perfectly fine. 
I'll change it but adds pointless verbosity, imho. :D

self-core-impl/src/main/java/com/selfxdsd/core/contracts/invoices/StoredInvoice.java
criske: Cumulates values from the whole stream, into a single value, basically a sum in our case.

First arg is the initial value

Second arg lambda is the reducer
`acc` is the accumlated value until that moment.
`curr` is the current value in stream.
The result is acc + rate * curr.timeSpentInHours, basically the new accumulated value.

The third argument lambda is used in parallel streams, it could be null in our case.

amihaiemil: @criske Thanks for explaining. 

I just like to keep it simple where possible. A simple sum method should not make you think: "wtf does reducing mean".

self-api/src/main/java/com/selfxdsd/api/Contract.java
amihaiemil: @criske Just out of principle. I cannot understand why Java allows public fields in the first place...

amihaiemil:
@rultor merge it

PR #173, opened 2020-06-07
StoredTask works with Contract

### LINKED ISSUE #172: StoredTask Contract Constructor
StoredTask should be constructed from:

a) the Project, if it is not assigned to anyone yet.
b) the Contract (which also holds the Project), if it is assigned.

amihaiemil:
Fixes #172 

A StoredTask can be created from a Project, if it's not assigned, or from a Contract, if there is one (if it is assigned).

Working with Contract simplifies the query for the assignee a lot, because the assignee is the owner of the Contract, so we don't have to query through the project's contributors and their contracts anymore.


amihaiemil:
@rultor merge it

Issue #164, opened 2020-06-06
Invoices.java:11-12: Implement and test InMemoryInvoices....


0pdd:
The puzzle `163-912d9b79` from #163 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/27bc74cf7e8c6ecdda356183621da6f6a9845ddf/self-api/src/main/java/com/selfxdsd/api/Invoices.java#L11-L12

The puzzle was created by cristianpela on 06-Jun-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
I'll work on this

amihaiemil:
paid with transaction ID 9RV72720Y86457822

0pdd:
The puzzle `163-912d9b79` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #174, opened 2020-06-07
#164 Implemented and partially tested InMemoryInvoices.

criske:
PR for #164 

Left puzzle to continue testing it and do some refactoring.

### DIFF
--- a/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
+++ b/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
@@ -6,7 +6,8 @@ import com.selfxdsd.core.contracts.invoices.StoredInvoice;
 
 import java.time.Duration;
 import java.util.*;
-import java.util.stream.Collectors;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 /**
@@ -169,7 +170,7 @@ public final class InMemoryInvoices implements Invoices {
     }
 
     /**
-     * Invoices of a contract.
+     * Invoices of contract.
      */
     private static final class ContractInvoices implements Invoices {
         /**
@@ -184,8 +185,13 @@ public final class InMemoryInvoices implements Invoices {
 
         /**
          * Stored invoices for this contract.
+         * We use a stream because we don't want to load in
+         * memory all the contract invoices.
+         * <br>
+         * In order to "reuse" them, since streams are one time use only,
+         * we wrap the stream in a supplier, .
          */
-        private final List<Invoice> contractInvoices;
+        private final Supplier<Stream<Invoice>> contractInvoices;
 
         /**
          * Current active invoice.
@@ -202,10 +208,9 @@ public final class InMemoryInvoices implements Invoices {
                                  final Storage storage) {
             this.contractId = contractId;
             this.storage = storage;
-            this.contractInvoices = new ArrayList<>(StreamSupport
+            this.contractInvoices = () ->StreamSupport
                 .stream(storage.invoices().spliterator(), false)
-                .filter(i -> i.contractId().equals(contractId))
-                .collect(Collectors.toUnmodifiableList()));
+                .filter(i -> i.contractId().equals(contractId));
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


criske:
Sorry if the PR is too big.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * Active contract invoice id.
+     */
+    private final Map<Contract.Id, Integer> activeIds = new HashMap<>();
+
+    /**
+     * Payed contract invoices ids.
+     */
+    private final Map<Contract.Id, List<Integer>> payedIds = new HashMap<>();
```
amihaiemil, line 46: @criske I'm not sure we will need all these 4 maps. I'd like to model it so that we have only one map representing teh table, like other InMeory* implementations.

But let's leave it as it is for now, there will probably be changes anyway, later on.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
```diff
@@ -0,0 +1,277 @@
...
+        final Contract.Id contractId = foundActiveEntry.getKey();
+        final Integer invoiceId = foundActiveEntry.getValue();
+        payedIds.computeIfAbsent(contractId, k -> new ArrayList<>())
+            .add(invoiceId);
+        activeIds.remove(contractId);
+    }
+
+    /**
+     * Contract Invoices delegate. Manages invoices for a contract.
+     */
```
amihaiemil, line 174: @criske Can you just say "Invoices of a Contract", to follow the pattern of other similar classes?

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
```diff
@@ -0,0 +1,277 @@
...
+
+        /**
+         * Stored invoices for this contract.
+         * We use a stream because we don't want to load in
+         * memory all the contract invoices.
+         * <br>
+         * In order to "reuse" them, since streams are one time use only,
+         * we wrap the stream in a supplier, .
+         */
+        private final Supplier<Stream<Invoice>> contractInvoices;
```
amihaiemil, line 194: @criske Why do we need supplier of Stream? Can't we work with a List (actual filtering being done in an upper layer), like other examples such as ProjectTasks, ProjectContributors etc?

criske: If a contract has 1 million invoices then all of them will be loaded in mem, if we're using a list. Basically we will throw away the paging benefits from self-storage.
The stream will not emit any items until you "subscribe" to it.

criske: Maybe it is a little bit forced in this case.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
```diff
@@ -184,8 +185,13 @@ public void pay(final int id) {
...
         /**
          * Stored invoices for this contract.
+         * We use a stream because we don't want to load in
+         * memory all the contract invoices.
+         * <br>
+         * In order to "reuse" them, since streams are one time use only,
+         * we wrap the stream in a supplier, .
          */
-        private final List<Invoice> contractInvoices;
+        private final Supplier<Stream<Invoice>> contractInvoices;
```
criske, line 194: Reversed back to stream.
If a new **invoice** is created in storage, the list will not be updated, unless we refresh by doing the whole "collect" ceremony from constructor.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
amihaiemil: @criske But there will not be 1M invoices in the Contract, for sure. Same for Tasks. A Project cannot have more than, say 1000 tasks.

Think of each Task (and Invoice element) as of a Github issue. How many Github Projects do you know, which have 1M Issues? :D

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
amihaiemil: @criske Ok, let's leave it like that. Maybe I didn't completely understand the whole picture :D

amihaiemil:
@rultor merge it

Issue #175, opened 2020-06-07
InMemoryInvoices.java:19-23: Refactor and test...


0pdd:
The puzzle `165-63222a61` from #165 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/101eccca529220e334e10b7bae5db6e8c070d447/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java#L19-L23

The puzzle was created by cristianpela on 07-Jun-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
I'll work on this.
I'll focus on testing and leave refactoring part for another puzzle, if that's ok.

0pdd:
The puzzle `165-63222a61` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

amihaiemil:
paid with transaction id 1BD17495F00273448

PR #181, opened 2020-06-08
#175 - Finished testing InMemoryInvoices.

criske:
PR for #175 

Left puzzle to refactor and test ContractInvoices.

### DIFF
--- a/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
+++ b/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
@@ -16,11 +16,8 @@ import java.util.stream.StreamSupport;
  * @author criske
  * @version $Id$
  * @since 0.0.3
- * @todo #165:30min Refactor and test ContractInvoices to be part
- *  of self-impl API .
- *  Also InvoiceTask should have its own concrete implementation in self-impl.
- *  Use that in InMemoryInvoices.
- *  Continue testing InMemoryInvoices.
+ * @todo #181:30min Refactor and test ContractInvoices to be part
+ *  of self-impl API.
  */
 public final class InMemoryInvoices implements Invoices {
 
@@ -194,11 +191,6 @@ public final class InMemoryInvoices implements Invoices {
         private final Supplier<Stream<Invoice>> contractInvoices;
 
         /**
-         * Current active invoice.
-         */
-        private Integer activeId;
-
-        /**
          * Ctor.
          *
          * @param contractId Contract id
@@ -244,18 +236,33 @@ public final class InMemoryInvoices implements Invoices {
                 throw new IllegalStateException("Task is not part of"
                    + " the contract.");
             }
-            InvoiceTask invoiceTask = storage.invoices().add(task, timeSpent);
-            activeId = invoiceTask.invoiceId();
-            return invoiceTask;
+            return storage.invoices().add(task, timeSpent);
         }
 
         @Override
         public boolean isPaid(final int id) {
-            return activeId != id;
+            checkInvoiceOfContract(id);
+            return storage.invoices().isPaid(id);
         }
 
         @Override
         public List<InvoiceTask> tasks(final int id) {
+            checkInvoiceOfContract(id);
+            return storage.invoices().tasks(id);
+        }
+
+        @Override
+        public Iterator<Invoice> iterator() {
+            return contractInvoices.get().iterator();
+        }
+
+        /**
+         * Checks if the invoice is part of the current contract.
+         * @param id Invoice id.
+         * @throws IllegalStateException when invoice is not part
+         * of the contract.
+         */
+        private void checkInvoiceOfContract(final int id) {
             final Invoice invoice = contractInvoices
                 .get()
                 .filter(i -> i.invoiceId() == id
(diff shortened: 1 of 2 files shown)


self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoicesTestCase.java
```diff
@@ -134,33 +138,153 @@ public void paysInvoice(){
...
+            Contract.Roles.ARCH
+        );
+        storage.invoices().add(this.mockAssignedTask(
+            "123", storage, project, contributor, Contract.Roles.DEV),
+            Duration.ofHours(1));
+        storage.invoices().add(this.mockAssignedTask(
+            "124", storage, project, contributor, Contract.Roles.ARCH),
+            Duration.ofHours(2));
 
+        storage.invoices().ofContract(contractIdTwo).isPaid(1);
```
amihaiemil, line 238: @criske What is the ``1`` argument in ``.isPaid(1)``? Is it the InvoiceId?

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoicesTestCase.java
```diff
@@ -134,33 +138,153 @@ public void paysInvoice(){
...
+            contributor.username(),
+            contributor.provider(),
+            Contract.Roles.ARCH
+        );
+        storage.invoices().add(this.mockAssignedTask(
+            "123", storage, project, contributor, Contract.Roles.DEV),
+            Duration.ofHours(1));
+        storage.invoices().add(this.mockAssignedTask(
+            "124", storage, project, contributor, Contract.Roles.ARCH),
+            Duration.ofHours(2));
```
amihaiemil, line 284: @criske Shouldn't this test fail here or when adding the first task above? Why does it fail only when trying to read the Task from teh contract (the last line  bellow)?

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
```diff
@@ -266,12 +273,6 @@ public boolean isPaid(final int id) {
                 throw new IllegalStateException("Invoice with id"
                     + id + " is not part of the contract.");
             }
-            return storage.invoices().tasks(id);
-        }
-
-        @Override
-        public Iterator<Invoice> iterator() {
-            return contractInvoices.get().iterator();
         }
```
amihaiemil, line 276: @criske Why remove the iterator?

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
```diff
@@ -193,11 +190,6 @@ public void pay(final int id) {
          */
         private final Supplier<Stream<Invoice>> contractInvoices;
 
-        /**
-         * Current active invoice.
-         */
-        private Integer activeId;
```
amihaiemil, line 199: @criske Why remove this?

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoicesTestCase.java
criske: Yes

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoicesTestCase.java
criske: Will not fail because we adding two tasks each linked with a different contract. (one contract has DEV role other ARCH, diff keys basically).

The last line will fail becase we're trying to get tasks of an invoice that is not part of the contract.
Invoice with id 1 is part of the contractOne not contractTwo.

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
criske: Is not removed, it will fail the build if so; was just displaced I guess.
https://github.com/self-xdsd/self-core/blob/b23b0cf6e44fa481ce403a926a44de8bb251cdac/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryInvoices.java
criske: Was a bug with that aproach.
Will fail the test when we check if the invoice contract is paid. You have to check the actual storage if payment is done.
Here is the corner case :
`Invoices invoicesOfContract = contract.invoices() .ofContract(contractId);`
then we add contract task through overall invoices not through **invoicesOfContract**:
`storage.invoices().add(taskOfContract, duration)`
then we "pay"
`storage.invoices().pay(contractInvoiceId)`
Now if you call  `invoicesOfContract.isPaid(contractInvoiceId) ` the old code will return **false** instead of **true** (stale state of **activeId**).

amihaiemil:
@criske thanks for the answers!

amihaiemil:
@rultor merge

Issue #201, opened 2020-06-11
Implement Tasks :: Contract.tasks()


amihaiemil:
It should return the Tasks assigned to the Contract.

criske:
I'll work on this.

amihaiemil:
@criske sure, go ahead

amihaiemil:
paid with transaction id: 5C873822XL412951G

PR #202, opened 2020-06-11
#201 Started implementing and testing Contract#tasks.

criske:
PR for #201

Left puzzle to continue implementing and testing ContractTasks.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
@@ -82,8 +82,12 @@ public final class ContractTasks implements Tasks {
         }else{
             //@checkstyle BooleanExpressionComplexity (10 lines)
             final List<Task> ofContract = StreamSupport
-                .stream(this.storage.tasks()
-                    .ofContract(id).spliterator(), false)
+                .stream(this.storage.tasks().spliterator(), false)
+                .filter(t -> t.assignee() != null && t.assignee()
+                    .username().equals(id.getContributorUsername())
+                    && t.project().repoFullName().equals(id.getRepoFullName())
+                    && t.project().provider().equals(id.getProvider())
+                    && t.role().equals(id.getRole()))
                 .collect(Collectors.toList());
             tasks = new ContractTasks(id, ofContract, this.storage);
         }
--- a/self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
@@ -9,6 +9,7 @@ import com.selfxdsd.api.storage.Storage;
 import java.util.Iterator;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 /**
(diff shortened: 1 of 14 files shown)


criske:
sorry, wrong commit message. it should be "test coverage fixes".

self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public Tasks ofContract(final Contract.Id id) {
+        Tasks tasks;
+        if(this.contractId.equals(id)){
+            tasks = this;
+        }else{
+            //@checkstyle BooleanExpressionComplexity (10 lines)
+            final List<Task> ofContract = this.tasks.stream()
```
criske, line 83: This logic is wrong here. I'll fix asap.

self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
```diff
@@ -80,12 +81,9 @@ public Tasks ofContract(final Contract.Id id) {
...
-            final List<Task> ofContract = this.tasks.stream()
-                .filter(t -> t.assignee() != null && t.assignee()
-                    .username().equals(id.getContributorUsername())
-                    && t.project().repoFullName().equals(id.getRepoFullName())
-                    && t.project().provider().equals(id.getProvider())
-                    && t.role().equals(id.getRole()))
+            final List<Task> ofContract = StreamSupport
+                .stream(this.storage.tasks()
+                    .ofContract(id).spliterator(), false)
                 .collect(Collectors.toList());
```
criske, line 87: Actually this will throw stackoverflow exception with real implementation of storage...

amihaiemil: @criske But you don't need all this logic here. Just throw ISE, like in the other examples and say: "These are the Tasks of Contract X, you cannot see other Contract's tasks here"

self-core-impl/src/main/java/com/selfxdsd/core/tasks/StoredTask.java
```diff
@@ -192,5 +192,10 @@ public String role() {
         public Invoices invoices() {
             return null;
         }
+
+        @Override
+        public Tasks tasks() {
+            return null;
+        }
```
amihaiemil, line 199: @criske Why do we need this method inside ``StoredTask``?

criske: that's from Unassigned Contract.

self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
criske: You're right :D

self-core-impl/src/main/java/com/selfxdsd/core/tasks/StoredTask.java
amihaiemil: @criske Right, forgot about that one...

amihaiemil:
@rultor merge it

Issue #203, opened 2020-06-11
Contract.java:33-35: Finish implementing and testing...


0pdd:
The puzzle `204-4a2ef409` from #204 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/a9a1a6c2d13883f574b10fd1656ef0eb33e3fda5/self-api/src/main/java/com/selfxdsd/api/Contract.java#L33-L35

The puzzle was created by cristianpela on 11-Jun-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
I'll work on this

amihaiemil:
@criske sure, go ahead

amihaiemil:
paid with transaction id: 2UG77083V2195670R

0pdd:
The puzzle `204-4a2ef409` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #204, opened 2020-06-12
#203 Finished implementing and testing ContractTasks.

criske:
PR for #203 

Left puzzle to implement and test ofContract() in ProjectTasks, ContributorTasks and InMemoryTasks using ContractTasks.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
@@ -8,6 +8,7 @@ import com.selfxdsd.api.storage.Storage;
 
 import java.util.Iterator;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Active tasks of a Contract. This class <b>just represents</b>
@@ -16,6 +17,8 @@ import java.util.List;
  * @author criske
  * @version $Id$
  * @since 0.0.6
+ * @todo: #205:30min Implement and test ofContract() in
+ *  ProjectTasks, ContributorTasks and InMemoryTasks using ContractTasks.
  */
 public final class ContractTasks implements Tasks {
     /**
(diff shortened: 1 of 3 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
```diff
@@ -63,13 +71,24 @@ public Task register(final Issue issue) {
...
     public Tasks ofContributor(final String username,
                                final String provider) {
-        throw new UnsupportedOperationException("Not implemented yet");
+        final List<Task> ofContributor = this.tasks.stream()
+            .filter(t -> t.assignee() != null
+                && t.assignee().username().equals(username)
+                && t.assignee().provider().equals(provider))
+            .collect(Collectors.toList());
+        return new ContributorTasks(username, provider, ofContributor,
+            this.storage);
```
amihaiemil, line 91: @criske Since these are already the Tasks of a Contract, I think in this method you should just check if the Contributor ID (username + prov) are matching the Contributor of the Contract. If yes, return ``this``, otherwise throw an exception ("Already seeing the Tasks of Contributor X, can't find other Tasks here").

Makes sense? :D

self-core-impl/src/main/java/com/selfxdsd/core/tasks/ContractTasks.java
```diff
@@ -63,13 +71,24 @@ public Task register(final Issue issue) {
     @Override
     public Tasks ofProject(final String repoFullName,
                            final String repoProvider) {
-        throw new UnsupportedOperationException("Not implemented yet");
+        final List<Task> ofProject = this.tasks.stream()
+            .filter(t -> t.project().repoFullName().equals(repoFullName)
+                && t.project().provider().equals(repoProvider))
+            .collect(Collectors.toList());
+        return new ProjectTasks(repoFullName, repoProvider, ofProject,
+            this.storage);
```
amihaiemil, line 79: @criske Same as bellow, these are the Tasks of a Contract and the Contract belongs to a single Project.

amihaiemil:
@criske nice, thanks!

amihaiemil:
@rultor merge

amihaiemil:
@rultor try to merge again

PR #206, opened 2020-06-13
#205 SelfContract lazy-loading of Project and Contributor

### LINKED ISSUE #205: StoredContract lazy ctor
We have a constructor which expects the Project and Contributor (eager loading).
We should also have a ctor which doesn't expect these dependencies.

amihaiemil:
fixes #205 

If the Project and Contributor are not given as ctor parameters, they will be read from the Storage.


amihaiemil:
@rultor merge it

Issue #208, opened 2020-06-14
Implement equals/hashCode for StoredContributor and others


amihaiemil:
StoredContributor, StoredContract, StoredProject etc should implement equals and hashcode based on their IDs.

criske:
I'll work on this

amihaiemil:
sure, go ahead

amihaiemil:
@criske One thing: in equals, when you check for the object's type with ``instanceof``, use the Interface, not the Class. 

This is a mistake I find in most equals functions: they ask the objects to be of the same class instead of the same interface.

criske:
yep, that's what I did

amihaiemil:
paid with transaction id 7JC54856LK201074A

PR #211, opened 2020-06-16
#208 Implemented equals and hashCode contracts for Stored* classes.

criske:
PR for #208


self-core-impl/src/main/java/com/selfxdsd/core/contracts/invoices/StoredInvoice.java
```diff
@@ -113,4 +114,15 @@ public BigDecimal totalAmount() {
         return this.storage.invoices().ofContract(this.contractId)
             .tasks(this.id);
     }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(this.id);
```
criske, line 120: this might be redudant? this.id it's enough.

criske:
@amihaiemil sure

amihaiemil:
@rultor merge it

PR #215, opened 2020-06-21
Clean Invoices

### LINKED ISSUE #214: Fix Invoices Architecture
Right now the Invoices part does not respect the overall architecture (mainly because it works with a ``List<InvoiceTask>`` instead of the plural ``InvoicedTasks``). 

This causes the code to be much more complicated than it should be. Let's refactor it and make it respect the established architecture.

amihaiemil:
Fixes #214 

Cleaned the Invoices interface and left puzzle for continuing with ``InvoicedTasks``.


amihaiemil:
@rultor merge it

Issue #220, opened 2020-06-22
Task.java:32-34: Implement and write unit tests for...


0pdd:
The puzzle `216-e1db1cae` from #216 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/922cc789b80e9af2d5ad3d36b6a156b7434386d9/self-api/src/main/java/com/selfxdsd/api/Task.java#L32-L34

The puzzle was created by @amihaiemil on 22-Jun-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

0pdd:
The puzzle `216-e1db1cae` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #221, opened 2020-06-22
#220 Task.value() and Task.estimation()

amihaiemil:
PR for #220 

* Implemented and tested Task.value() and Task.estimation()
* Used the Task's actual value when registering it to an Invoice (InMemoryInvoicedTasks)


amihaiemil:
@rultor merge it

PR #231, opened 2020-06-27
#229 Self extends AutoCloseable + unit test

### LINKED ISSUE #229: Make Self Autocloseable
Since ``Self`` is the highest abstraction it should also be ``Autocloseable`` and close any resources it holds. 

For instance, it holds a reference to the ``Storage`` which is itself ``Autocloseable`` and should be closed when ``Self.close()`` is called.

amihaiemil:
Fixes #229 
Self is now AutoCloseable and closes the underlying Storage on close.


amihaiemil:
@rultor merge it

PR #240, opened 2020-06-29
#239 GithubIssue/s uses JsonResources + made classes package protected

### LINKED ISSUE #239: Use JsonResources (JdkHttp) in GithubIssues
We should use JsonResources in GithubIssues as well.

amihaiemil:
fixes #239 

* Made GithubIssue/s work with JsonResources instead of making direct HTTP calls
* Mage classes GithubIssues and GithubIssue package protected in the core package -- they should not be visible to the outside.


amihaiemil:
@rultor merge it

Issue #100, opened 2020-05-24
Issue.java:32-34: Implement Comments::Issue.comments()....


0pdd:
The puzzle `97-bcc117e9` from #97 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/fdbd1d044d9248d08980205b75dd71cc9e86f974/self-api/src/main/java/com/selfxdsd/api/Issue.java#L32-L34

The puzzle was created by @amihaiemil on 24-May-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
I'll work on this

amihaiemil:
@criske sure, go ahead (listing comments should happen via the iterator() method; i.e. interface Comments extends Iterable<Comment>)

amihaiemil:
paid with transaction id 57492179F0155882Y

PR #243, opened 2020-06-30
#100 Comment/Comments API.

criske:
PR for #100 

- added Comments::Issue.comments()
- created and tested GithubComment
- started implementing GithubIssueComments

Left puzzle to finish implementing and testing GithubIssueComments

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
@@ -1,0 +1,51 @@
+package com.selfxdsd.core;
+
+import com.selfxdsd.api.Comment;
+import com.selfxdsd.api.Comments;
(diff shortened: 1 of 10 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.net.URI;
+import java.util.Iterator;
+
+/**
+ * Github Issue Comments.
+ * @author criske
+ * @version $Id$
+ * @since 0.0.8
+ * @todo #244:30min Implement and test the remaining methods
```
amihaiemil, line 14: @criske the number here has to be ``#100``, since that is the parent ticket.

amihaiemil:
@rultor merge

Issue #244, opened 2020-06-30
GithubIssueComments.java:14-15: Implement and test the...


0pdd:
The puzzle `100-a80e16f6` from #100 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/0468b4566626fc97f7663b30f844e53dfce59b7e/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java#L14-L15

The puzzle was created by cristianpela on 30-Jun-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
I'll work on this.

waiting for #237

amihaiemil:
@criske sure, I'll take care of that asap

amihaiemil:
@criske I've implemented the POST method in #245 :)

criske:
@amihaiemil Nice, just one question: should I test using Grizzly, or just unit test it with mocks?

amihaiemil:
@criske You can write a todo for the unit tests. We'll write unit tests with a mock version of JsonResources. I'll implement that a bit later in #246. 

It doesn't really make sense to Grizzly tests, since the only valuable assertion would be URI -- JdkHttp post already has integration tests using Grizzly.

amihaiemil:
paid with transaction id: 0ER62346G31483402

0pdd:
The puzzle `100-a80e16f6` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #247, opened 2020-07-01
#244 Finished implementing and testing GithubIssueComments.

criske:
PR for #244 
Left puzzle for GithubIssueComments pagination.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
@@ -5,6 +5,7 @@ import com.selfxdsd.api.Comments;
 
 import javax.json.Json;
 import javax.json.JsonObject;
+import java.net.HttpURLConnection;
 import java.net.URI;
 import java.util.Iterator;
 import java.util.List;
@@ -17,8 +18,8 @@ import java.util.stream.Collectors;
  * @version $Id$
  * @since 0.0.8
  * @todo #100:30min Provide a pagination solution in iterator() using
- *  *  Github response headers, when {@link Resource} API has access to http
- *  *  response headers (issue #241).
+ *  Github response headers, when {@link Resource} API has access to http
+ *  response headers (issue #241).
  */
 public final class GithubIssueComments implements Comments {
 
@@ -50,13 +51,12 @@ public final class GithubIssueComments implements Comments {
     }
 
     @Override
-    public Comment post(final String body, final String accessToken) {
+    public Comment post(final String body) {
         final Resource resource = resources.post(
                 this.commentsUri,
                 Json.createObjectBuilder().add("body", body).build(),
-                accessToken
-        );
-        if (resource.statusCode() == 201) {
+                "");
+        if (resource.statusCode() == HttpURLConnection.HTTP_CREATED) {
             return new GithubComment(resource.asJsonObject());
         } else {
             throw new IllegalStateException("Github Issue Comment "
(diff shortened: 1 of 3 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
```diff
@@ -3,16 +3,22 @@
...
  * Github Issue Comments.
+ *
  * @author criske
  * @version $Id$
  * @since 0.0.8
- * @todo #100:30min Implement and test the remaining methods
- *  in GithubIssueComments.
+ * @todo #100:30min Provide a pagination solution in iterator() using
+ *  *  Github response headers, when {@link Resource} API has access to http
+ *  *  response headers (issue #241).
```
amihaiemil, line 21: @criske puzzle here is incorrect, line 2 and 3 have useless ``*``.

self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
```diff
@@ -28,24 +34,49 @@
...
+        }
+        this.commentsUri = URI.create(issueUriStr + slash + "comments/");
         this.resources = resources;
     }
 
     @Override
-    public Comment post(final String body) {
-        //POST /repos/:owner/:repo/issues/:issue_number/comments
-        throw new UnsupportedOperationException("Not implemented yet");
+    public Comment post(final String body, final String accessToken) {
```
amihaiemil, line 53: @criske This method should not have the accessToken as parameter.

Initially, I wanted the token to come from the ``Github`` parent: passed to ``GithubIssues`` in constructor, then passed to ``GithubIssue`` and finally to ``GithubComments``. 

But now, I see it would  be better if ``JdkHttp`` encapsulated the token as well. I will do that in #248. 

I suggest you leave this method unimplemented for now and write a ``todo`` for implementing it later.

self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
```diff
@@ -28,24 +34,49 @@
...
+        }
     }
 
     @Override
     public Iterator<Comment> iterator() {
-        //GET /repos/:owner/:repo/issues/:issue_number/comments
-        throw new UnsupportedOperationException("Not implemented yet");
+        final Resource resource = resources.get(this.commentsUri);
+        List<Comment> comments;
+        if (resource.statusCode() == 200) {
```
amihaiemil, line 71: @criske Can you please use ``HttpUrlConnection.STATUS_OK`` instead of hardcoding ``200``?

self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
```diff
@@ -28,24 +34,49 @@
...
+                    + "was not created");
+        }
     }
 
     @Override
     public Iterator<Comment> iterator() {
-        //GET /repos/:owner/:repo/issues/:issue_number/comments
-        throw new UnsupportedOperationException("Not implemented yet");
+        final Resource resource = resources.get(this.commentsUri);
+        List<Comment> comments;
```
amihaiemil, line 70: @criske This variable can be final :D

self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
criske: Yeah, didn't have access to accessToken, so I had to update the api, or go refactoring all over the project. It's way much better if  `JdkHttp` encapsulate the token.

Wouldn't be better if, after reverting to initial `post`, I just pass an empty string to `JsonResources#post`? The code is still correct plus tests don't rely on token anyway.

amihaiemil: @criske yes, that's also ok :)

amihaiemil:
@rultor merge it please

Issue #248, opened 2020-07-01
Implement JsonResources.withToken(...)


amihaiemil:
This method should give us an instance of JsonResources which has an API access token for authenticating the requests.

PR #250, opened 2020-07-02
#248 Authenticated JsonResources

amihaiemil:
PR for #248 

Implemented JsonResources :: JsonResources.authenticated(accessToken).
This way, JsonResources encapsulates the accessToken and doesn't expect it as method param, so
we will not have to pass around the token anymore.


amihaiemil:
@rultor merge it

PR #253, opened 2020-07-02
#252 JsonResources.patch implemented

amihaiemil:
PR for #252 

Implemented and tested the patch method.


amihaiemil:
@rultor merge it

Issue #122, opened 2020-05-27
Get user Org Repos


amihaiemil:
We should get the Org repos of the authenticated User, to which the User has Admin permissions.

Get Orgs: https://developer.github.com/v3/orgs/#list-your-organizations
Check Repo permissions: https://developer.github.com/v3/repos/collaborators/#review-a-users-permission-level

criske:
@amihaiemil I'll work on this

amihaiemil:
@criske This will be quite a lot of work. Start small, with ``Organizations :: provider.organizations()``. Organizations will be an Iterable of Organization and will represent all the orgs to which the User has admin rights.

amihaiemil:
paid with transaction id 62H57924UM3490632

PR #257, opened 2020-07-02
#122 Provider Organizations API scaffold.

criske:
PR for #122.
Left puzzle to start implementing Github organizations.


criske:
When this PR passes I'll continue the follow up pdd within the same bounty.
Just wanted to make sure that I had the right aproach.

self-api/src/main/java/com/selfxdsd/api/Organization.java
```diff
@@ -0,0 +1,64 @@
...
+     * description etc...
+     * @return Detail
+     */
+    Detail detail();
+
+    /**
+     * Contains extra information about the organization like name,
+     * description etc...
+     */
+    interface Detail {
```
amihaiemil, line 42: @criske Detail is not needed for now. The json() method is enough if we want some specific info about the Organization... we will be interested only in the repos, most likely.

self-api/src/main/java/com/selfxdsd/api/Organization.java
```diff
@@ -0,0 +1,64 @@
...
+     * Organization ID.
+     * @return String.
+     */
+    String organizationId();
+
+    /**
+     * Organization repos.
+     * @return OrganizationRepos.
+     */
+    OrganizationRepos repos();
```
amihaiemil, line 23: @criske Can you rename the interface ``OrganiztionRepos`` to simply ``Repos``? ``Repos`` is an interface which represents more repositories, but where these repositories are coming from is not important.

We will most likely have 2 main implementations of Repos: 1 for an Organization's repositories and 1 for a user's private repositories.

amihaiemil:
@criske When making a PR please, always specify the ticket you are solving in the description, so Github links the Issue with the PR automatically -- it's not enough to put it only in the title.

amihaiemil:
@criske looks good, thanks!

amihaiemil:
@rultor merge it

Issue #254, opened 2020-07-02
JsonResources.java:40-44: Add other methods such as...


0pdd:
The puzzle `252-820887a6` from #252 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/909001591f009487f18d5618e18ab2cda56f4cbe/self-core-impl/src/main/java/com/selfxdsd/core/JsonResources.java#L40-L44

The puzzle was created by amihaiemil on 02-Jul-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
I'll work on this

amihaiemil:
@criske if you want to start with the mock version, I think you should start with the get method.

It will encapsulate a ``Map<URI, JsonValue>`` which we will initialize in the constructor.

0pdd:
The puzzle `252-820887a6` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #260, opened 2020-07-04
#254 Implemented and tested MockJsonResources.

criske:
PR for #254.

Updated puzzle.

### DIFF
--- a/self-core-impl/src/test/java/com/selfxdsd/core/mock/MockJsonResources.java
+++ b/self-core-impl/src/test/java/com/selfxdsd/core/mock/MockJsonResources.java
@@ -1,0 +1,215 @@
+package com.selfxdsd.core.mock;
+
+import com.selfxdsd.core.JsonResources;
+import com.selfxdsd.core.Resource;
+
+import javax.json.Json;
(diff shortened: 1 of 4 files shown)


criske:
I've tried a different aproach, by using a callback that acts as a "controller".

self-core-impl/src/test/java/com/selfxdsd/core/mock/MockJsonResources.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.function.Function;
+
+/**
+ * A mock implementation of {@link JsonResources} used to unit test
+ * {@link com.selfxdsd.api.Provider} requests.
+ * @author criske
+ * @version $Id$
+ * @since 0.0.8
+ */
+public final class MockJsonResources implements JsonResources {
```
amihaiemil, line 20: @criske This class looks interesting, but it's not clear to me how it will be used. Can you update the JavaDoc here and write an example of usage? You can write code in JavaDocs between ``<pre>`` tags and indentation will be preserved.

amihaiemil:
@rultor good to merge

Issue #242, opened 2020-06-30
JsonResources, accessToken: passing an interface instead of string. 


criske:
Providers might use different headers for authenticated requests.
Right know JdkHttp is using github's auth header _Authentication_, but Gitlab, as far as I know, is using a different header: _Private-Token_.
So the interface AccessToken might have to methods to implement: 
- header : the header name (Authentication, Private-Token)
- value: the token

For example a GithubAccessToken would look like:
- header: `Authentication`
- value: `token <token>`

amihaiemil:
@criske It's a good idea, thanks :)

criske:
I'll work on this

PR #267, opened 2020-07-08
#242 AccessToken core-impl.

criske:
- Github and Gitlab AccessToken implemented and tested.

PR for #242

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/JsonResources.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/JsonResources.java
@@ -49,7 +49,7 @@ public interface JsonResources {
      * @param accessToken Access token.
      * @return JsonResources.
      */
-    JsonResources authenticated(final String accessToken);
+    JsonResources authenticated(final AccessToken accessToken);
 
     /**
      * Get the Resource at the specified URI.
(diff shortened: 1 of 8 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/JsonResources.java
```diff
@@ -214,15 +214,13 @@ private HttpRequest request(
...
+            if(this.accessToken != null) {
                 request = HttpRequest.newBuilder()
                     .uri(uri)
                     .method(method, body)
                     .header("Content-Type", "application/json")
-                    .header(
-                        "Authorization",
-                        "token " + this.accessToken
-                    ).build();
+                    .header(this.accessToken.header(), this.accessToken.value())
```
amihaiemil, line 222: @criske In the case of Github, you have to preppend the ``token `` word in front of the actual token's value. Maybe you can modify the ``value()`` method to return ``"token " + value``?

criske: @amihaiemil It's prepended in GithubAccessToken constructor.

amihaiemil:
@rultor merge it please

amihaiemil:
@criske there is a compilation error, apparently :D

amihaiemil:
@rultor merge

PR #271, opened 2020-07-08
#270 Project.resolve(), unit test, todo

### LINKED ISSUE #270: Project.resolve(json event, String token)
We should implement this method which will take the JSON payload received via provider webhooks, check if the secret token matchets and delegate the work further to the PM.

amihaiemil:
Fixes #270 

Started implementing Project.resolve()
Left puzzle to continue in a further ticket
Removed Project.webHookToken() because there's actually no need to expose it.


amihaiemil:
@rultor merge it

Issue #259, opened 2020-07-03
Organizations.java:8-9: Start implementing Github...


0pdd:
The puzzle `122-8525164d` from #122 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/4c219b0434b6a8b5b679f2180083f1d6bdd515df/self-api/src/main/java/com/selfxdsd/api/Organizations.java#L8-L9

The puzzle was created by cristianpela on 02-Jul-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
@amihaiemil Can I work on this?

amihaiemil:
@criske Yes, but don't write a lot of code, please. Start small, with listing the Organizations :D

0pdd:
The puzzle `122-8525164d` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #274, opened 2020-07-09
#259 Implemented and tested Github Org/Orgs.

criske:
- Left puzzle to implement and test Github Org repos.

PR for #259

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GithubOrganization.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GithubOrganization.java
@@ -11,9 +11,10 @@ import javax.json.JsonObject;
  * @author criske
  * @version $Id$
  * @since 0.0.9
- * @todo #122:30 Implement and test GithubOrganization Repos.
+ * @todo #122:30 Continue to implement and test Github Organization Repos for
+ *  the authenticated User.
  */
-public final class GithubOrganization implements Organization {
+final class GithubOrganization implements Organization {
 
     /**
      * The Organization in JSON format as returned by Github's API.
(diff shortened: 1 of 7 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/GithubOrganization.java
```diff
@@ -0,0 +1,54 @@
...
+
+import javax.json.JsonObject;
+
+/**
+ * A Github Provider Organization.
+ *
+ * @author criske
+ * @version $Id$
+ * @since 0.0.9
+ * @todo #122:30 Implement and test GithubOrganization Repos.
```
amihaiemil, line 14: @criske Can you specify more details in this todo? 0pdd will throw an exception if the todo is too short. Usually, it has to have 10-15 words.

self-core-impl/src/main/java/com/selfxdsd/core/GithubOrganization.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public final class GithubOrganization implements Organization {
```
amihaiemil, line 16: @criske Make this class package-protected pls. It should not be visible outside of the core.

self-core-impl/src/main/java/com/selfxdsd/core/GithubOrganizations.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Iterator;
+
+/**
+ * A Github Provider Organizations.
+ *
+ * @author criske
+ * @version $Id$
+ * @since 0.0.9
+ */
+public final class GithubOrganizations implements Organizations {
```
amihaiemil, line 19: @criske This class also package-protected pls.

amihaiemil:
@rultor merge it please

criske:
Weird, initially the checkstyle passed on my end.

amihaiemil:
@rultor merge

Issue #275, opened 2020-07-09
GithubOrganization.java:14-15: Continue to implement and...


0pdd:
The puzzle `122-8be42be5` from #122 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/da7cd6e5975c7e6947df2cd4d9a4984a1eee470d/self-core-impl/src/main/java/com/selfxdsd/core/GithubOrganization.java#L14-L15

The puzzle was created by cristianpela on 09-Jul-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
I'll continue to work on this.

0pdd:
The puzzle `122-8be42be5` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #276, opened 2020-07-09
#275 Implemented and tested GithubOrganizationRepos.

criske:
PR for #275

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GithubOrganizationRepos.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GithubOrganizationRepos.java
@@ -1,0 +1,106 @@
+package com.selfxdsd.core;
+
+import com.selfxdsd.api.Repo;
+import com.selfxdsd.api.Repos;
+import com.selfxdsd.api.User;
+import com.selfxdsd.api.storage.Storage;
+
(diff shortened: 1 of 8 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/GithubOrganizationRepos.java
```diff
@@ -0,0 +1,106 @@
...
+     *
+     * @param repoData Repo as JSON.
+     * @return Repo.
+     */
+    private Repo buildRepo(final JsonValue repoData) {
+        final URI repoUri = URI.create(((JsonObject) repoData)
+            .getString("url"));
+        return new GithubRepo(this.resources,
+            repoUri,
+            this.owner,
```
criske, line 103: @amihaiemil I think here, instead of "this.owner" should be the owner from json repoData, which is not necessarily the authenticated user?.

amihaiemil: @criske well, the "owner" of an organization repo is the Organization itself.
From our perspective, the owner is the authenticated User because he has admin rights on the repo :D

amihaiemil:
@rultor merge it

PR #279, opened 2020-07-09
Event interface, Issues.received, Comments.received

### LINKED ISSUE #278: Introduce Event (type, Issue, Comment)
This interface will represent an Event received by Self, which is sent by the Provider. 

We should also have the possibility of creating an Issue and a Comment from an existing JSON (in case we receive it in the Event).

amihaiemil:
Fixes #278 

Added the Event interface which represents an Event sent to Self via one of the webhooks.
Added Issues.received to create an Issue from a JsonObject (a webhook event usually has the Issue in json).
Added Comments.received to create a Comment from a JsonObject (a webhook event can contain a comment in json)
Added unit tests
Changed Project.resolve(JsonObject,...) to Project.resolve(Event,...);

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Comments.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Comments.java
@@ -1,5 +1,7 @@
 package com.selfxdsd.api;
 
+import javax.json.JsonObject;
+
 /**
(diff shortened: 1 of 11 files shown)


amihaiemil:
@rultor merge it

Issue #277, opened 2020-07-09
User.java:30-31: Allow {@link User} interface to have...


0pdd:
The puzzle `277-e4f1192a` from #277 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/e7388d374744703552f38187140311f9e3b17777/self-api/src/main/java/com/selfxdsd/api/User.java#L30-L31

The puzzle was created by cristianpela on 09-Jul-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
@amihaiemil I'll work on this

amihaiemil:
@criske sure, go ahead :)

PR #282, opened 2020-07-10
#277 Added Organizations to User API.

criske:
- implemented and tested(where needed) organizations in User implementations.
- moved organizations uri logic from Github to GithubOrganizations.

Left puzzle to start implementing Gitlab organizations (groups).

PR for #277

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/StoredUser.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/StoredUser.java
@@ -22,11 +22,13 @@
  */
 package com.selfxdsd.core;
 
+import com.selfxdsd.api.Organizations;
 import com.selfxdsd.api.Projects;
 import com.selfxdsd.api.Provider;
 import com.selfxdsd.api.User;
 import com.selfxdsd.api.storage.Storage;
 
+import java.net.URI;
 import java.util.Objects;
 
 /**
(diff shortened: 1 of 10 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/StoredUser.java
```diff
@@ -102,6 +112,24 @@ public Projects projects() {
...
 
+    @Override
+    public Organizations organizations() {
+        final Organizations organizations;
+        switch (this.provider){
+            case Provider.Names.GITHUB: organizations =
+                new GithubOrganizations(this.resources,
+                    URI.create("https://api.github.com"), this,
+                    this.storage);
+                break;
```
criske, line 123: @amihaiemil Having Provider.Names as enum that encapsulates the name and provider Uri would be really handy :).
If not just class of constants (eg: Provider.Uri) would do it too. Right now Github uri is copy/pasted in a bunch of places.

criske:
@amihaiemil Tests for `BaseSelf.Authenticated`, `PmUser` are missing. Should I write them in this PR or open new issue?

self-core-impl/src/main/java/com/selfxdsd/core/StoredUser.java
amihaiemil: @criske I see, but let's leave it as it is for now. We'll make these kind of small imrovements later :D

amihaiemil:
@criske 

> Tests for BaseSelf.Authenticated, PmUser are missing. Should I write them in this PR or open new issue?

Please open another issue :)

amihaiemil:
@rultor merge it

Issue #285, opened 2020-07-11
User.java:30-31: Implement and test Organizations for...


0pdd:
The puzzle `278-d86e619b` from #278 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/9f26118eeffed11be9c8da66dfb3fba71332ce0b/self-api/src/main/java/com/selfxdsd/api/User.java#L30-L31

The puzzle was created by cristianpela on 10-Jul-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
@amihaiemil I'll work on this. I'll start with GitlabOrganizations.

amihaiemil:
@criske sure, go ahead :D

criske:
@amihaiemil Since this issue has bounty, I'll countinue with Gitlab Org Repos within the same bounty. Current PR #297 was pretty much a bunch of copy/paste from Github Orgs. :D

0pdd:
The puzzle `278-d86e619b` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

amihaiemil:
paid with transaction id: 5397001380987720Y

PR #297, opened 2020-07-18
#285 GitlabOrganizations/GitlabOrganization.

criske:
- implemented and tested.
- updated puzzle.

PR for #285


self-api/src/main/java/com/selfxdsd/api/User.java
```diff
@@ -27,8 +27,8 @@
  * @author Mihai Andronache (amihaiemil@gmail.com)
  * @version $Id$
  * @since 0.0.1
- * @todo #278:30min Implement and test Organizations for Gitlab
- *  provider.
+ * @todo #278:30min Continue with GitlabOrganizations by implementing and
```
amihaiemil, line 30: @criske YOu have to also change the parent ticket number, otherwise the todo will not close, I think :D

amihaiemil:
@rultor merge it please

Issue #298, opened 2020-07-18
User.java:30-31: Continue with GitlabOrganizations by...


0pdd:
The puzzle `298-42a404da` from #298 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/dd2648290b4baad63189299c60deb8f2e33daac2/self-api/src/main/java/com/selfxdsd/api/User.java#L30-L31

The puzzle was created by cristianpela on 18-Jul-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

amihaiemil:
paid with transaction id 7WV57429PP6023543

amihaiemil:
@criske why is still still open, didn't you remove the puzzle from the code? :D

amihaiemil:
@criske yes, I saw you removed it... strange.

PR #299, opened 2020-07-18
#298 Implemented GitlabOrganizationRepos.

criske:
- implemented and tested.
- left puzzle for gitlab invitations.

PR for #298

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GitlabOrganizationRepos.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GitlabOrganizationRepos.java
@@ -1,0 +1,94 @@
+package com.selfxdsd.core;
+
+import com.selfxdsd.api.Repo;
+import com.selfxdsd.api.Repos;
+import com.selfxdsd.api.User;
+import com.selfxdsd.api.storage.Storage;
+
+import javax.json.JsonArray;
+import javax.json.JsonObject;
+import javax.json.JsonValue;
+import java.net.HttpURLConnection;
+import java.net.URI;
+import java.util.Iterator;
+
+/**
+ * A Gitlab Provider Organization Repos.
+ *
+ * @author criske
+ * @version $Id$
+ * @since 0.0.9
+ */
+final class GitlabOrganizationRepos implements Repos {
+
+    /**
+     * Organization Repos URI owned by current User.
+     */
+    private final URI uri;
+
+    /**
+     * Current authenticated User.
+     */
+    private final User owner;
+
+    /**
+     * Gitlab's JSON Resources.
+     */
+    private final JsonResources resources;
+
+    /**
+     * Storage used by Organization Repo.
+     */
+    private final Storage storage;
+
+    /**
+     * Ctor.
+     * @param organizationId Organization Id.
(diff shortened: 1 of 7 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/GitlabOrganizationRepos.java
```diff
@@ -0,0 +1,94 @@
...
+     * @param owner Current authenticated User.
+     * @param resources Gitlab's JSON Resources.
+     * @param storage Storage used by Organization Repo.
+     */
+    GitlabOrganizationRepos(final String organizationId,
+                            final User owner,
+                            final JsonResources resources,
+                            final Storage storage) {
+        this.uri = URI.create("https://gitlab.com/api/v4/groups/"
+            + organizationId + "/projects?simple=true&owned=true");
```
amihaiemil, line 56: @criske the URI should be constructed on the fly, starting from ``GitLab`` and going down.
See how Github is implemented :D 

Or is there a special reason why you did it like this?

self-core-impl/src/main/java/com/selfxdsd/core/GitlabRepo.java
```diff
@@ -54,6 +54,24 @@
...
+     * @return GitlabRepo.
+     */
+    static GitlabRepo createFromName(final String repoName,
+                                     final JsonResources resources,
+                                     final User owner,
+                                     final Storage storage) {
+        final URI repo = URI.create("https://gitlab.com/api/v4/projects/"
+            + repoName);
+        return new GitlabRepo(resources, repo, owner, storage);
+    }
```
amihaiemil, line 73: @criske Why is this method needed?

self-core-impl/src/main/java/com/selfxdsd/core/GitlabOrganizationRepos.java
criske: Is easier to test. For Github makes sense cause there we dealing with org repos uri extracted from json.
https://github.com/self-xdsd/self-core/blob/dd2648290b4baad63189299c60deb8f2e33daac2/self-core-impl/src/test/java/com/selfxdsd/core/GithubOrganizationTestCase.java#L53
But here we only need the org `id` and with this aproach I'll avoid testing the uri twice: in` GitlabOrganizationTestCase` and in `GithubOrganizationReposTestCase`; But only once see: `GitlabOrganizationReposTestCase#fetchesAndFilterOk` in PR

self-core-impl/src/main/java/com/selfxdsd/core/GitlabRepo.java
criske: Same logic for uri creation is used in different places (`Gitlab` and `GitlabOrganizationRepos`). But the main inconvenient is that Gitlab API doesn't give full api-url for repos in their json responses as Github does. So each time we need to extract the repo name from json's `path`-key  and from there build the Uri used by `GitlabRepo`.

amihaiemil:
@criske I see, ok.

amihaiemil:
@rultor merge it please

Issue #317, opened 2020-07-26
Invoice.java:13-15: An Invoice should have the method...


0pdd:
The puzzle `315-3e2d84bd` from #315 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/b46c3bff05a7aaee8d51691f4250856dbf8e3721/self-api/src/main/java/com/selfxdsd/api/Invoice.java#L13-L15

The puzzle was created by @amihaiemil on 26-Jul-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

0pdd:
The puzzle `315-3e2d84bd` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #318, opened 2020-07-26
#317 Invoice.register(Task) + unit tests

amihaiemil:
PR for #317 
An Invoice can now register a Task. It will also check if the Task belongs to the same Contract or if it is paid already (it will complain in both situatons). 
Added unit tests too.


amihaiemil:
@rultor merge it

PR #342, opened 2020-08-05
#333 Provider.repo(name) -> Provider.repo(owner, name)

### LINKED ISSUE #333: Implement Provider.repo(owner, name)
At the moment we only have ``Provider.repo(name)`` which takes the repo's name and appends it to the username of the authenticated user. 

We also need a method to dictate the user (especially in the case of org repos, where the "owner" is the Organization).

Read [this](https://github.com/self-xdsd/self-core#cash-bounties) for details about the bounty.

amihaiemil:
Fixes #333 
The method now accepts the owner as well - it was not correct to assume that the owner was always the authenticated User, since the owner can also be an Organization.

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Provider.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Provider.java
@@ -38,12 +38,12 @@ public interface Provider {
     String name();
 
     /**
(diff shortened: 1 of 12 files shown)


amihaiemil:
@rultor merge

Issue #341, opened 2020-08-04
Exposing paging to storage api.


criske:
Though there is some paging done internally in `self-storage`, any listing request from `self-web` will load all the results in memory regardless. I think we should be able to select pages through `self-storage` something like `xxxPaged = storage.xxx().page(page)`. Adapting paging is straightforward, as long as we're using streams instead of passing lists or iterables for different `self-core` storage resources implementations. That's because there are places where we can't access the sql queries and have to rely on streams' skip/limit operators.

The `Paged` interface will have enough info for clients to figure out what page they can select (from 1 to totalPages)
```java
public interface Paged {
    Page current();
    int totalPages();
}
```
(`Page` is just a data class that encapsulates page number and size)

Let's take an example for paging `Projects`:

The call would look like:
`ProjectsPaged projects = storage.projects().page(new Page(1, 10))`
```java
interface ProjectsPaged extends Projects, Paged {}
```
The `ProjectsPage` implementation is simple and will be used by any implementations of `Projects`
```java
class ProjectsPagedImpl implements ProjectsPaged {

    private final Page current;
    private final int totalRecords;
    private final Projects projects;

    ProjectsPagedImpl(final Page current,
                      final int totalRecords,
                      final Projects projects) {
     //...
   }

   @Override
   public Page current() {
       return this.current;
   }

   @Override
   public int totalPages() {
       return this.totalRecords / this.current.getSize();
   }

    @Override
    public ProjectsPaged page(final Page page) {
        throw new IllegalStateException("Already called page");
    }

  // delegate other methods to this.projects
}
```
Here is how paging for PmProjects would look like:
```java
class PmProjects implements Projects {
    //...
    @Override
    public ProjectsPaged page(final Page page) {
        final Supplier<Stream<Project>> pagedProjects = () ->
            this.projects.get()
                .skip((page.getNumber() - 1) * page.getSize())
                .limit(page.getSize());
        return new ProjectsPagedImpl(page,
            this.totalRecords,
            new PmProjects(this.pmId, this.totalRecords, pageProjects));
    }
}
```

amihaiemil:
@criske Sounds good, can you start with a proof of concept? :D

criske:
@amihaiemil sure, I'll do one for `Projects`.

amihaiemil:
paid with transaction id 1D4908014F2773708

PR #345, opened 2020-08-06
#341 Storage Paged api & Projects paging api.

criske:
- added Storage Paged API.
- paging for Projects by updating the API and implemented DefaultProjectsPaged.
- updated InMemoryProjects, PmProjects and UserProjects to use streams and paging.
- fixed InMemoryContractsTestCase#addThrowsExceptionWhenContributorNotFound not fully mocked - was failing after updating Projects implementations to use streams.

Left two puzzle to write unit tests for paging: one for DefaultProjectsPaged and other for InMemoryProjects, PmProjects and UserProjects.

PR for #341


self-api/src/main/java/com/selfxdsd/api/ProjectsPaged.java
```diff
@@ -0,0 +1,12 @@
...
+
+import com.selfxdsd.api.storage.Paged;
+
+/**
+ * Paginated version of {@link Projects}.
+ * @author criske
+ * @version $Id$
+ * @since 0.0.13
+ */
+public interface ProjectsPaged extends Projects, Paged {
```
amihaiemil, line 11: @criske Do we really need this new interface? I would make it like this: ``Projects extends Paged<Project>``. And then add an abstract class called ``PagedProjects``, which would be extended by ``StoredProjects``. Wouldn't that be possible?

self-api/src/main/java/com/selfxdsd/api/storage/Paged.java
```diff
@@ -0,0 +1,66 @@
...
+         * Page size.
+         */
+        private final int size;
+
+        /**
+         * Ctor.
+         * @param number Page number.
+         * @param size Page size.
+         */
+        public Page(final int number, final int size) {
```
amihaiemil, line 44: @criske A page is actually defined by ``index`` and ``size`` -- index where we start the page and size of the page. Maybe your ``number`` is actually the start index?

self-core-impl/src/main/java/com/selfxdsd/core/projects/UserProjects.java
```diff
@@ -51,17 +52,17 @@
     /**
      * The projects.
      */
-    private final List<Project> projects;
+    private final Supplier<Stream<Project>> projects;
```
amihaiemil, line 55: @criske Is it possible to make these List -> Supplier changes in a separate PR?

self-api/src/main/java/com/selfxdsd/api/Projects.java
```diff
@@ -69,4 +71,12 @@ Project getProjectById(
         final String repoFullName,
         final String repoProvider
     );
+
+    /**
+     * Get the Projects at the provided Page.
+     * @param page Page number.
+     * @return ProjectsPaged
+     */
+    ProjectsPaged page(final Paged.Page page);
```
amihaiemil, line 80: @criske I think ``Projects page(final Paged.Page page)`` would make more sense :D

self-api/src/main/java/com/selfxdsd/api/ProjectsPaged.java
criske: @amihaiemil Sure we could do that. With my aproach we could list projects either paged either all at once. But I guess, in hindsight, having an endpoint to all projects is useless in our case :D . I'll change the design and see if it works.

self-api/src/main/java/com/selfxdsd/api/storage/Paged.java
criske: @amihaiemil the page number is the page number starting from 1.
Having a table index based aproach will require the client to know/guess the "table" size imo.
For example if we have 100 records and call `/projects/2/10` then paged implemention will know that it must fetch 10 projects starting from index 10, instead of letting the client to do that logic otherwise.

self-core-impl/src/main/java/com/selfxdsd/core/projects/UserProjects.java
criske: @amihaiemil since this was a proof of concept I've thought that is better to reflect paging across all Projects implementations. Anyway I think if we're going to use `PagedProjects` as abstract class, as you sugested, these implementations will extend that class and will be forced to use streams.

self-api/src/main/java/com/selfxdsd/api/Projects.java
criske: @amihaiemil in this draft `ProjectsPaged` is needed because the frontend needs the Paged "header" in order to construct the ui paging links from 1 to totalpages and highlight the current page.

amihaiemil:
@rultor merge it please

criske:
@amihaiemil there was a small bug InMemoryProjects where projects map not using the map passed from constructor. Dunno if rultor was using my fixed commit to merge. Anyway unit tests will catch it most likely.

Issue #290, opened 2020-07-11
Write unit tests for other User implementations.


criske:
Write tests for `BaseSelf.Authenticated` and `PmUser` in order to increase test coverage.

ANadrowski:
@amihaiemil I'd like to work on this. Could you assign it to me?

amihaiemil:
@ANadrowski sure, go ahead :D

amihaiemil:
@ANadrowski FYI, I put a bounty on this ticket : D

ANadrowski:
@amihaiemil It was a busy week here, but I will take a look asap :D

amihaiemil:
@ANadrowski no worries, whenever you have time :D

ANadrowski:
Hi @amihaiemil  
I'm sorry for the delay here. 
Feel free to remove the bounty but if possible let it still assigned to me and I will try to solve it in the next days (probably next weekend)... thanks.

amihaiemil:
@ANadrowski no problem, whenever you have time 👍

ANadrowski:
@amihaiemil thanks :)

ANadrowski:
Hi @amihaiemil @criske could you give me a tip here?

A good test for _BaseSelf.Authenticated_ should be to create an instance of it, passing the values in the constructor (User and token) and test its methods (_username()_, _email()_, _provider()_ and _projects()_)?

... and the same for _PmUser_?

I observed that the _BaseSelf.Authenticated_ is used in the test _platformLoginWorks()_, but there it is just tested if _Self_ is an instance of BaseSelf.Authenticated. So I think that the new test should test its methods. 

Am I right?

Thank you

PS. I'm sorry for this dumb question...

amihaiemil:
@ANadrowski Yes, a good unit test usually means you instantiate the object and then make assertions on the methods.

But the key is to mock the dependencies. For instance, you want to see that ``Authenticated.username()`` returns the username but you are not interested in the ``username()`` method from the ``User``. So you mock the User.

Here is an example: 

```java
final User user = Mockito.mock(User.class);
Mockito.when(user.username()).thenReturn("john");

final Authenticated authenticated = new Authenticated(user, "tok3n");
MatcherAssert.assertThat(
    authenticated.username(),
    Matchers.equalsTo("john")
);
```

ANadrowski:
@amihaiemil that is awesome, thanks! :D

ANadrowski:
@amihaiemil 
[...]

PR #346, opened 2020-08-06
Adds unit test for Authenticated and adds puzzle for PmUser

ANadrowski:
PR for #290 
Adds unit test for Authenticated and adds puzzle for PmUser.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
@@ -38,6 +38,8 @@ import java.util.UUID;
  * @since 0.0.1
  * @todo #212:30min Test equals and hashcode methods for each Stored*
  *  classes (StoredProjectManager, StoredProject etc.).
+ *
+ * @todo #290:15min Create unit test for PmUser class.
  */
 public final class StoredProjectManager implements ProjectManager {
 
--- a/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
@@ -39,7 +39,8 @@ import java.util.UUID;
  * @todo #212:30min Test equals and hashcode methods for each Stored*
  *  classes (StoredProjectManager, StoredProject etc.).
  *
- * @todo #290:15min Create unit test for PmUser class.
+ * @todo #290:15min Unit tests have been added for class Authenticated. 
+ *  We should also add unit tests for class PmUser.
  */
 public final class StoredProjectManager implements ProjectManager {
 
@@ -213,6 +214,5 @@ public final class StoredProjectManager implements ProjectManager {
         public Projects projects() {
             return this.manager.projects();
         }
-
     }
 }
--- a/self-core-impl/src/test/java/com/selfxdsd/core/SelfCoreTestCase.java
+++ b/self-core-impl/src/test/java/com/selfxdsd/core/SelfCoreTestCase.java
@@ -24,11 +24,9 @@ package com.selfxdsd.core;
 
 import com.selfxdsd.api.*;
 import com.selfxdsd.api.storage.Storage;
-import com.selfxdsd.core.BaseSelf.Authenticated;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.Test;
-import static org.mockito.Matchers.anyString;
 import org.mockito.Mockito;
 
 /**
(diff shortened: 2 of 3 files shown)


self-core-impl/src/test/java/com/selfxdsd/core/SelfCoreTestCase.java
```diff
@@ -107,5 +111,33 @@ public void closesStorageOnClose() throws Exception {
...
+        MatcherAssert.assertThat(
+            authenticated.email(),
+            Matchers.equalTo("john@john.com")
+        );
+        Mockito.verify(provider).withToken(anyString());
+        MatcherAssert.assertThat(
+            authenticated.projects(),
+            Matchers.equalTo(all)
+        );
+    }
```
amihaiemil, line 142: @ANadrowski This test is overall OK, but it tests the whole ``Authenticated`` class (all of its methods). A unit test should cover only one method. 

Can you please make a new test class, called ``AuthenticatedTestCase`` and add 4 tests? One for username, one for email, one for provider/withTocken and one for projects? 

Just beak down this big test into 4 smaller ones. For instance, in the case of ``Authenticated.username()`` you do not have to mock any ``Projects``.

self-core-impl/src/test/java/com/selfxdsd/core/SelfCoreTestCase.java
```diff
@@ -107,5 +111,33 @@ public void closesStorageOnClose() throws Exception {
...
 
+        MatcherAssert.assertThat(
+            authenticated.username(),
+            Matchers.equalTo("john")
+        );
+        MatcherAssert.assertThat(
+            authenticated.email(),
+            Matchers.equalTo("john@john.com")
+        );
+        Mockito.verify(provider).withToken(anyString());
```
amihaiemil, line 137: @ANadrowski This line verifies that ``provider.withToken(...)`` has been called once with any String. I think you can make it even more precise, like this:

```java
Mockito.verify(provider).withToken("tok3n");
```

This way you are sure that ``withToken`` has been called with the initial token.

self-core-impl/src/test/java/com/selfxdsd/core/SelfCoreTestCase.java
```diff
@@ -107,5 +111,33 @@ public void closesStorageOnClose() throws Exception {
...
+    public void authenticatedUser() {
+        final User user = Mockito.mock(User.class);
+        Mockito.when(user.username()).thenReturn("john");
+        Mockito.when(user.email()).thenReturn("john@john.com");
+        final Provider provider = Mockito.mock(Provider.class);
+        Mockito.when(user.provider()).thenReturn(provider);
+        final Projects all = Mockito.mock(Projects.class);
+        Mockito.when(user.projects()).thenReturn(all);
+        final Authenticated authenticated = new Authenticated(user, "tok3n");
+        final Provider prov = new Authenticated(user, "tok3n").provider();
```
amihaiemil, line 127: @ANadrowski You don't need to make a new instance of ``Authenticated`` here, just use the one above :D

self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
```diff
@@ -38,6 +38,8 @@
  * @since 0.0.1
  * @todo #212:30min Test equals and hashcode methods for each Stored*
  *  classes (StoredProjectManager, StoredProject etc.).
+ *
+ * @todo #290:15min Create unit test for PmUser class.
```
amihaiemil, line 42: @ANadrowski This todo needs to have more words (10-15 words), otherwise 0pdd will complain. How about something like this: 

```
@todo #290:15min Unit tests have been added for class Authenticated. We should also
 add unit tests for class PmUser.
```

self-core-impl/src/test/java/com/selfxdsd/core/SelfCoreTestCase.java
ANadrowski: @amihaiemil yeah, it makes sense. I'll implement it :)

self-core-impl/src/test/java/com/selfxdsd/core/SelfCoreTestCase.java
ANadrowski: ok :)

self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
ANadrowski: ok :)

self-core-impl/src/test/java/com/selfxdsd/core/SelfCoreTestCase.java
ANadrowski: @amihaiemil I did not understand this. Do you have in mind something like this: `final Provider prov = authenticated.provider();` ?

amihaiemil: @ANadrowski yes, exactly :D

ANadrowski: @amihaiemil thanks :)

ANadrowski:
@amihaiemil I just sent the requested changes, please take a look when possible :D

amihaiemil:
@ANadrowski very nice, thank you!

amihaiemil:
@rultor merge it

ANadrowski:
@amihaiemil thank tou!

Issue #380, opened 2020-08-11
ProjectContributors.java:47-50: Once we have the Project...


0pdd:
The puzzle `376-eb663e41` from #376 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/355d899cd35a74275e252214f74c26bbe52f065f/self-core-impl/src/main/java/com/selfxdsd/core/contributors/ProjectContributors.java#L47-L50

The puzzle was created by @amihaiemil on 11-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

0pdd:
The puzzle `376-eb663e41` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #383, opened 2020-08-11
#380 Elect only affordable contributors

amihaiemil:
PR for #380 
When electing a Contributor, we must also make sure that the final price of the Task goes into the available budget of the Project.
Added unit tests.
Left puzzle to implement Wallet.pay(...)


amihaiemil:
@rultor merge it

PR #399, opened 2020-08-14
#398 StoredContract.value() adds up the PM's commission too

### LINKED ISSUE #398: Addapt SelfContract.value() to add up PM commission for each Task
For each Task, we should also add the PM's commission.

amihaiemil:
fixes #398


amihaiemil:
@rultor merge

PR #401, opened 2020-08-14
#396 Electing uses PM's commission too.

criske:
PR for #396

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/contributors/ProjectContributors.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/contributors/ProjectContributors.java
@@ -40,8 +40,6 @@ import java.util.stream.Stream;
  * @author Mihai Andronache (amihaiemil@gmail.com)
  * @version $Id$
  * @since 0.0.4
- * @todo #393:30min Take the PM's commission into account when electing
- *  a Contributor (to make sure we can afford to pay everything).
  */
 public final class ProjectContributors implements Contributors {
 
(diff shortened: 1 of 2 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/contributors/ProjectContributors.java
```diff
@@ -180,6 +178,10 @@ public Contributor elect(final Task task) {
                                 ).divide(
                                     BigDecimal.valueOf(60),
                                     RoundingMode.HALF_UP
+                                ).add(task
```
criske, line 181: @amihaiemil the above if condition shouldn't be: `(contract.role().equals(task.role()) && contract.project().equals(task.project())` ?

amihaiemil: @criske yes, we could add that check too. Can you write a puzzle for it? 
Also, in this case, I think it's easier if you say:

```
.add(
    this.project.projectManager().commission()
)
```

self-core-impl/src/test/java/com/selfxdsd/core/contributors/ProjectContributorsTestCase.java
```diff
@@ -494,6 +517,11 @@ private Project mockProject(
...
         Mockito.when(project.wallet()).thenReturn(wallet);
 
+        final ProjectManager projectManager = Mockito
+            .mock(ProjectManager.class);
+        Mockito.when(projectManager.commission()).thenReturn(pmCommission);
+        Mockito.when(project.projectManager()).thenReturn(projectManager);
+
         return project;
     }
 }
```
amihaiemil, line 527: @criske can you add a unit test where all the contributors have an affordable price, but it goes our of range because of the commission? 

For example, Wallet has 15000 available cents, the Contributor's price is 14500, and the commission is 600. Then the total price (15100) is our of rance :D

amihaiemil:
@rultor merge it

PR #405, opened 2020-08-15
#403 Invoice.register() now accepts the commission as well

### LINKED ISSUE #403: StoredInvoice.java:14-15: Addapt method register to also...
The puzzle `400-bb6375ab` from #400 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/a4fc78c81619535123409e72eab512d1ef253b08/self-core-impl/src/main/java/com/selfxdsd/core/contracts/invoices/StoredInvoice.java#L14-L15

The puzzle was created by amihaiemil on 15-Aug-20. 

amihaiemil:
PR for #403


amihaiemil:
@rultor merge it

Issue #404, opened 2020-08-15
ProjectContributors.java:43-44: When electing a...


0pdd:
The puzzle `402-0cf23d35` from #402 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/278603527b12bdbfaa7345110a99c3d578e154e1/self-core-impl/src/main/java/com/selfxdsd/core/contributors/ProjectContributors.java#L43-L44

The puzzle was created by cristianpela on 15-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

amihaiemil:
@criske wanna take care of it? By the way, I think it's easier if you just use the encapsulated Project (no need to read the Project from the contributor's Contract every time).

criske:
@amihaiemil sure

amihaiemil:
paid with transaction id 8VA15175A0094282N

0pdd:
The puzzle `402-0cf23d35` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #406, opened 2020-08-15
#404 Checks ProjectContributors project eq. Task's project.

criske:
PR for #404

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/contributors/ProjectContributors.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/contributors/ProjectContributors.java
@@ -152,15 +152,17 @@ public final class ProjectContributors implements Contributors {
      *
      * In the future, we might take more factors into account.
      * @param task Task requiring an assignee.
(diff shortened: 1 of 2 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/contributors/ProjectContributors.java
```diff
@@ -154,12 +152,16 @@ public Contributors ofProject(
...
-     * @return Contributor or null if nobody is found.
+     * @return Contributor or null if nobody is found or Task's Project
+     * not matching ProjectContributors Project.
      * @checkstyle ReturnCount (30 lines)
      * @checkstyle Indentation (30 lines)
      */
     @Override
     public Contributor elect(final Task task) {
+        if (!this.project.equals(task.project())) {
+            return null;
```
amihaiemil, line 163: @criske I think we should thrown an ISE in this case, since it's not the correct flow -- the given Task should always correspond to the Project.

amihaiemil:
@rultor merge it

PR #411, opened 2020-08-16
#410 StoredInvoice encapsulates its tasks

amihaiemil:
PR for #410 
Also left puzzle to remove the first constructor, because it's not actually needed.


amihaiemil:
@rultor merge it please

PR #422, opened 2020-08-17
#420 StoredTask.issue gets the PM's Issue

### LINKED ISSUE #420: StoredTask.java:38-40: Method issue() here should fetch...
The puzzle `418-e7efccc1` from #418 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/f9689a97eb0324d0e76eb33e5cb35de345a2aa4f/self-core-impl/src/main/java/com/selfxdsd/core/tasks/StoredTask.java#L38-L40

The puzzle was created by @amihaiemil on 17-Aug-20. 

amihaiemil:
Fixes #420 
puzzle to refactor some tests.


amihaiemil:
@rultor merge it please

PR #432, opened 2020-08-17
#430 Tasks.assign(Task, Contract, days) implemented and tested

amihaiemil:
PR for #430 

Method implemented in all implementations of ``Tasks``. Maybe some of those implementations should do some additional checks (to maintain the context), but let's leave that for later.


amihaiemil:
@rultor merge it

Issue #431, opened 2020-08-17
StoredTask.java:42-43: Implement and test method...


0pdd:
The puzzle `427-4eee3dd1` from #427 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/c938941bf5779db98cb131d18cc99439bfb42ddb/self-core-impl/src/main/java/com/selfxdsd/core/tasks/StoredTask.java#L42-L43

The puzzle was created by @amihaiemil on 17-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

criske:
@amihaiemil I'll work on this

amihaiemil:
@criske sure, go ahead

amihaiemil:
paid with transaction id 3V459057YM807335L

0pdd:
The puzzle `427-4eee3dd1` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

Issue #437, opened 2020-08-18
Add method Task.issueId()


amihaiemil:
We should have this accessor method, since we have this id in the Database. It makes no sense to call ``task.issue()`` when we only need the issue's ID.

PR #439, opened 2020-08-18
#431. Implemented+tested Task::unassign().

criske:
- added the corespondent method in Tasks Api.
- implemented and tested unassign() for filter Tasks classes.
- left puzzle to test InMemoryTasks#unassign()

PR for #431

### DIFF
--- a/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryTasks.java
+++ b/self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryTasks.java
@@ -130,6 +130,24 @@ public final class InMemoryTasks implements Tasks {
     }
 
     @Override
+    public Task unassign(final Task task) {
+        final TaskKey key = new TaskKey(
+            task.issue().issueId(),
+            task.project().repoFullName(),
+            task.project().provider()
+        );
+        final Task unassigned = new StoredTask(
+            task.project(),
+            key.issueId,
+            task.role(),
+            task.estimation(),
+            this.storage
+        );
+        this.tasks.put(key, unassigned);
+        return unassigned;
+    }
+
+    @Override
     public Tasks ofProject(final String repoFullName,
                            final String repoProvider) {
         final Supplier<Stream<Task>> tasksOf = () -> tasks.values()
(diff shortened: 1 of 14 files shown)


self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryTasksTestCase.java
```diff
@@ -37,6 +37,8 @@
  * @todo #135:30min Finish 'getTasksOfContributor()' test case
  *  when API permits that; as in when a Task can be assigned to a
  *  Contributor.
+ * @todo #439:30min Test assign and unassign Task using the in-memory
+ *  implementation of {@link Storage}.
```
amihaiemil, line 41: @criske You don't really need to use the in-memory storage mock. You can test Task.unassign(...) by just using Mockito :D

criske: @amihaiemil I've thought it would be nice to write it as an "integration" test, now that we have almost all in memory pieces implemented. But sure with mocks will be enough.
Now that #437 is implemented it's easier to write the test. Assign/unassign is using Task#issue() to get the id, which is a network call. Having issue id available will make things much better.
Should I update with the #437 PR and write the test, or just it leave as a puzzle?

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryTasks.java
```diff
@@ -129,6 +129,24 @@ public Task assign(
         return assigned;
     }
 
+    @Override
+    public Task unassign(final Task task) {
```
criske, line 133: @amihaiemil InMemoryTasks#unassign is implemented here :)

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryTasksTestCase.java
amihaiemil: @criske Just leave it as a puzzle :D

self-core-impl/src/test/java/com/selfxdsd/core/mock/InMemoryTasks.java
amihaiemil: @criske right, I missed it, sorry :D

amihaiemil:
@criske 

> now that we have almost all in memory pieces implemented.

Well, actually the InMemory infrastructure is beginning to be a bit useless, in my opinion. It was good in the beginning, but now the project caught a pretty good shape and we already have the real storage module implemented.

amihaiemil:
@criske can I merge this, or you still need to add a puzzle for testing? :D

criske:
@amihaiemil go ahead:)

amihaiemil:
@rultor merge it

PR #440, opened 2020-08-18
#437 StoredTask.issueId() implmeneted and tested

### LINKED ISSUE #437: Add method Task.issueId()

amihaiemil:
Fixes #437


amihaiemil:
@rultor merge it

Issue #451, opened 2020-08-19
Understand.java:39-41: Add the Confused implementation of...


0pdd:
The puzzle `446-a770e22e` from #446 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/399f64deab4d501e34ed2a9a95a6eb45e022210c/self-core-impl/src/main/java/com/selfxdsd/core/managers/Understand.java#L39-L41

The puzzle was created by amihaiemil on 19-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

amihaiemil:
@criske Wanna take this one? It's not complicated, just see how the Hello conversation is implemented.
The only difference is, the Confused conversation sill be the last one in the chain, so it won't encapsulate any ``next`` Conversation. 

The ``Confused`` conversation should simply build and return a SendReply step, with a reply message saying something like ``"@author Sorry, I did not understand that"``

criske:
@amihaiemil sure

amihaiemil:
paid with transaction id 23193855XX8306448

0pdd:
The puzzle `446-a770e22e` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #456, opened 2020-08-19
#451. Confused implemented + tested.

criske:
PR for #451

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/managers/Confused.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/managers/Confused.java
@@ -26,10 +26,7 @@ public final class Confused implements Conversation {
         final Step steps;
         if (Event.Type.CONFUSED.equals(event.type())) {
             steps = new SendReply(
-                String.format(
-                    event.project().language().reply("misunderstand.comment"),
-                    event.comment().author()
-                ),
+                event.project().language().reply("misunderstand.comment"),
                 lastly -> LOG.debug("Finished conversation.")
             );
         } else {
--- a/self-core-impl/src/main/java/com/selfxdsd/core/managers/Confused.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/managers/Confused.java
@@ -1,0 +1,42 @@
+package com.selfxdsd.core.managers;
+
+import com.selfxdsd.api.Event;
+import com.selfxdsd.api.pm.Conversation;
+import com.selfxdsd.api.pm.Step;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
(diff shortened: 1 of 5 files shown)


self-core-impl/src/test/resources/responses_en.properties
```diff
@@ -14,4 +14,5 @@ taskAssigned.comment=@%s this is your task now, please go ahead. Deadline (when
                      Estimation here is ``%s minutes``, that's how much you will be paid.\n\n\
                      Remember, you don't have to solve everything in this ticket - you can solve it partially\
                      and leave ``todo`` markers in the code, which will become future tasks.\n\n\
-                     If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.
\ No newline at end of file
+                     If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.
+misunderstand.comment=Didn't understand.
```
amihaiemil, line 18: @criske Can you add ``@%s`` in the beginning here? Like: ``@%s didn't understand that, sorry.``

self-core-impl/src/main/java/com/selfxdsd/core/managers/Confused.java
```diff
@@ -0,0 +1,39 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(
+        Confused.class
+    );
+
+    @Override
+    public Step start(final Event event) {
+        final Step steps;
+        if (Event.Type.CONFUSED.equals(event.type())) {
+            steps = new SendReply(
+                event.project().language().reply("misunderstand.comment"),
```
amihaiemil, line 29: @criske Can you format this reply to add the comment's author? Like [here](https://github.com/self-xdsd/self-core/blob/master/self-core-impl/src/main/java/com/selfxdsd/core/managers/Hello.java#L64).

String.format(...) will replace the ``%s`` with the specified value.

self-core-impl/src/test/java/com/selfxdsd/core/managers/ConfusedTestCase.java
```diff
@@ -0,0 +1,90 @@
...
+        MatcherAssert.assertThat(
+            step,
+            Matchers.allOf(
+                Matchers.notNullValue(),
+                Matchers.instanceOf(SendReply.class)
+            )
+        );
+
+        Mockito.verify(comments).post("> " + "gibberish" + "\n\n"
+            + "Didn't understand.");
```
amihaiemil, line 76: @criske you will also need to update this assert after you format the reply with the author :D

amihaiemil (changes requested):
@criske looks nice! Just a few comments :D

amihaiemil:
@rultor merge it please

Issue #466, opened 2020-08-20
Implement Deregister conversation


amihaiemil:
In this conversation, if the event type is "deregister",  the PM should unassign the Task and remove it from the DB. This will allow the project to have open Issues which are not tracked by the PM. 

Only the PO and ARCH roles should be able to give this command.

amihaiemil:
@criske Wanna start this one? The first thing we need is a sketch of Deregister and a Step which verifies the role of the author.

See how the Resign conversation and the AuthorIsAssignee check are implemented :D

criske:
@amihaiemil ok

criske:
@amihaiemil there is a problem, we don't have a way to remove a task from db.

amihaiemil:
paid with transaction id 9SM62392XE1717021

PR #467, opened 2020-08-20
#466. Deregister Conversation mostly implemented.

criske:
- implemented and tested AuthorHasRoles step.
- implemented Deregister conversation.
- mostly implemented RemoveTask step.
- left puzzle to test Deregister,RemoveTask + finish it.
- left puzzle to remove a task from db.

PR for #466

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/managers/AuthorHasRoles.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/managers/AuthorHasRoles.java
@@ -1,0 +1,101 @@
+/**
+ * Copyright (c) 2020, Self XDSD Contributors
+ * All rights reserved.
+ * <p>
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"),
+ * to read the Software only. Permission is hereby NOT GRANTED to use, copy,
+ * modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software.
+ * <p>
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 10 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/managers/AuthorHasRoles.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public void perform(final Event event) {
+        final Project project = event.project();
+        final String author = event.comment().author();
+        final Contributor contributor = project
+            .contributors()
+            .getById(author, project.provider());
+        if (contributor == null) {
```
criske, line 76: is `project.owner()` implicit a contributor? 
Else I should check to see if author is project.owner() and has the role.

amihaiemil: @criske The owner is not by default a contributor. I think we will make him a Contributor with role ``PO`` (Project Owner), but it's irrelevant for the code here.

When we register a Project, we could call ``Contributors.register(...)`` to add him as a contributor. So don't worry.

self-core-impl/src/main/java/com/selfxdsd/core/managers/Deregister.java
```diff
@@ -0,0 +1,97 @@
...
+import com.selfxdsd.api.pm.Conversation;
+import com.selfxdsd.api.pm.Step;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Conversation where a PM will unassign the Task and remove it from DB.
+ * @author criske
+ * @version $Id$
+ * @since 0.0.20g
```
amihaiemil, line 37: @criske typo here, can you remove the ``g``?

amihaiemil:
@rultor merge it

Issue #464, opened 2020-08-20
Implement Status conversation


amihaiemil:
When a contributor says "status" in an Issue, the PM should reply with current info about the Task (assignee, deadline etc).

PR #469, opened 2020-08-21
#464 Status conversation implemented and tested

amihaiemil:
PR for #464 

When a user says "status" the PM will reply with 3 possibilities: 
1. The Task is not registered
2. The Task is registered but not yet assigned to anyone
3. The Task is assigned to X since Y, until Z, with estimation E.


amihaiemil:
@rultor merge it

Issue #479, opened 2020-08-24
Repo.activate() should check if the repo is already active


amihaiemil:
If we already have a Project registered for that Repo, it should throw ``RepoAlreadyActiveException`` (runtime exception).

criske:
@amihaiemil I'll work on this

amihaiemil:
paid with transaction id 9PF80659YG762591G

PR #483, opened 2020-08-24
#479. Repo#activate() throws RepoAlreadyActiveException if active.

criske:
PR for #479

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/exceptions/RepoAlreadyActiveException.java
+++ b/self-api/src/main/java/com/selfxdsd/api/exceptions/RepoAlreadyActiveException.java
@@ -15,7 +15,7 @@ public final class RepoAlreadyActiveException extends RuntimeException {
      * @param repo Repo in question.
      */
     public RepoAlreadyActiveException(final Repo repo){
-        super("Rep \"" + repo.fullName() + "\" is already active.");
+        super("Repo \"" + repo.fullName() + "\" is already active.");
     }
 
 }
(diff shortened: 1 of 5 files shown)


self-api/src/main/java/com/selfxdsd/api/exceptions/RepoAlreadyActiveException.java
```diff
@@ -0,0 +1,21 @@
...
+ * @since 0.0.20
+ */
+public final class RepoAlreadyActiveException extends RuntimeException {
+
+    /**
+     * Ctor.
+     * @param repo Repo in question.
+     */
+    public RepoAlreadyActiveException(final Repo repo){
+        super("Repo \"" + repo.fullName() + "\" is already active.");
```
criske, line 18: I've tried to keep the same pattern as `InvoiceAlreadyPaidException`.
But in this case `repo.fullName()` will require another json request. So unless we find a way to cache these json responses (etags or something) -- or at least part of them like `fullName`, maybe we should pass the `fullName` string instead. (and possibly the `Provider` name, just to make the message clearer.)

amihaiemil: @criske Method [BaseRepo.json()](https://github.com/self-xdsd/self-core/blob/master/self-core-impl/src/main/java/com/selfxdsd/core/BaseRepo.java#L91) caches the json so there will be only one initial real call made, right? 

Or am I missing something? :D

amihaiemil (commented):
@criske looks nice, see my reply :D

self-api/src/main/java/com/selfxdsd/api/exceptions/RepoAlreadyActiveException.java
criske: You're right, dunno why I had the impression it didn't lol. You can merge it

amihaiemil:
@rultor merge it please

PR #484, opened 2020-08-25
#480 RepoNotFoundException implemented and used

### LINKED ISSUE #480: Throw RepoNotFoundException
If we receive other status than 200 OK when trying to fetch the repos data, we should throw RepoNotFoundException (runtime exception).

amihaiemil:
Fixes #480


amihaiemil:
@rultor merge it

PR #486, opened 2020-08-25
#470 Register scaffolding, AuthorHasRoles ANY + unit tests

### LINKED ISSUE #470: Implement the Register conversation
In this conversation, the user asks the PM to register an Issue as a Task. 

New Issues are registered as tasks automatically, but an Issue can be deregistered and then the user might wanna register it again. There is also the case of old Issues (which were opened before the PM joined the Project), which are not being reigistered automatically.

amihaiemil:
Fixes #470

Started implementing the Register conversation. Left puzzle to continue.
Addapted AuthorHasRoles to return true if the role is ``ANY``.
Unit tests.


amihaiemil:
@rultor merge it

Issue #489, opened 2020-08-25
Register.java:39-41: Implement an intermediary step where...


0pdd:
The puzzle `478-71d5692f` from #478 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/7fd0dcd083c04a58c0a23a53b72fc8eeb226fc3a/self-core-impl/src/main/java/com/selfxdsd/core/managers/Register.java#L39-L41

The puzzle was created by amihaiemil on 25-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

0pdd:
The puzzle `478-71d5692f` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #490, opened 2020-08-25
#489 RegisterIssue implemented, tested and used

amihaiemil:
PR for #489 

Implemented and tested step "RegisterIssue" and finished composition of steps inside Register.


amihaiemil:
@rultor merge it please

Issue #491, opened 2020-08-25
SendReply.java:38-42: Add a simple constructor which only...


0pdd:
The puzzle `489-893054ff` from #489 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/78aef3236b2b5689817007a4148699efec974fd5/self-core-impl/src/main/java/com/selfxdsd/core/managers/SendReply.java#L38-L42

The puzzle was created by amihaiemil on 25-Aug-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

0pdd:
The puzzle `489-893054ff` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #492, opened 2020-08-25
#491 Simple ctor for SendReply, to ease its usage 

amihaiemil:
PR for #491 

Added a simple String ctor for SendReply, to clean up the composition where SendReply is being used. Most of the times, SendReply is the last step in any conversation, so a final step logging "Conversation ended" is enough.


amihaiemil:
@rultor merge it

Issue #498, opened 2020-08-29
Register a Resignation


amihaiemil:
We need a way to register a resignation.

criske:
@amihaiemil I'll work on this

PR #505, opened 2020-08-30
#498. Implementend+tested Resignations#register.

criske:
PR for #498

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Task.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Task.java
@@ -30,7 +30,7 @@ import java.time.LocalDateTime;
  * @author Mihai Andronache (amihaiemil@gmail.com)
  * @version $Id$
  * @since 0.0.1
- * @todo #502:30min. Task api should expose resignations(), for easier access
+ * @todo #498:30min. Task api should expose resignations(), for easier access
  *  in ProjectContributors#elect(Task) algorithm, in order to filter out
  *  previously resigned contributors from Task.
  */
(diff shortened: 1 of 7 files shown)


criske:
@amihaiemil Wouldn't be better if `Tasks#unassign(Task)` call internally `Resignations#register` too? Now `Tasks#unassign(Task, Reason)` would return `Resignation` instead of just `Task`.

self-api/src/main/java/com/selfxdsd/api/Task.java
```diff
@@ -30,6 +30,9 @@
  * @author Mihai Andronache (amihaiemil@gmail.com)
  * @version $Id$
  * @since 0.0.1
+ * @todo #502:30min. Task api should expose resignations(), for easier access
```
amihaiemil, line 33: @criske 498 is the parent ticket here, not 502

amihaiemil (commented):
@criske just one tiny comment

amihaiemil:
@criske 

> Wouldn't be better if Tasks#unassign(Task) call internally Resignations#register too

I was also thinking about that option, but I'm not sure. I think maybe it's more clear if we keep them separate.
Plus, resignation should not always be registered -- for example, if the Task is deregistered (removed) by the PO or ARC, then we should not register a resignation.

criske:
@amihaiemil 
Calling  `Resignations#register` and `Tasks#unassign(Task)` sequentially might create problems imo  -- I think this in the order when execting a possible ResignStep.
These two should be executed in a transaction at the db level (supposing there are two tables `tasks` and `resignation_tasks`).

amihaiemil:
@criske yeah, they should theoretically be executed in the same transaction. But I think it's a bit overkill. It's not a big problem if registering the resignation fails.

So Task.unassign + Resignations.register, sequencially in this order should be fine... at least for now.

amihaiemil:
@rultor merge it please

PR #507, opened 2020-08-30
#503 Project reads wallets and active wallet from storage

amihaiemil:
PR for #503 

Added method Project.wallets() which reads the wallets from the storage.
Modified method Project.waller() to return the active wallet.


amihaiemil:
@rultor merge it please

PR #515, opened 2020-08-31
#514 Wallets.register(...) expects String identifier

### LINKED ISSUE #514: Wallets.register(...) should also accept the ID
It should also accept a String identifier of the Wallet in question.

amihaiemil:
Fixes #514


amihaiemil:
@rultor merge it please

PR #518, opened 2020-09-04
#516 Removed permission param when inviting Repo Collaborator

### LINKED ISSUE #516: Inviting user to Repo returns 422
422 UNPROCESSABLE ENTITY -- seems like Github cannot handle our request.

amihaiemil:
Fixes #516 
The ``permission`` parameter is only applicable for orgs -- if it's a user repo, the request will fail with ``422 UNPROCESSABLE ENTITY``. Since the default permission level is ``push`` (which is good enough), we removed this parameter.


amihaiemil:
@rultor merge it please

Issue #541, opened 2020-09-10
StripePayoutMethod.java:33-36: Implement Stripe API call...


0pdd:
The puzzle `538-6c60ba7a` from #538 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/cbd8ce58c751f167ef33b138183b0ceeff8e2e17/self-core-impl/src/main/java/com/selfxdsd/core/contributors/StripePayoutMethod.java#L33-L36

The puzzle was created by @amihaiemil on 10-Sep-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

0pdd:
The puzzle `538-6c60ba7a` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #543, opened 2020-09-10
#541 PayoutMethod.json() implemented and tested

amihaiemil:
PR for #541 

Instead of offering just an accessor method for the ``onboarding`` flag, we decided to offer an accessor method for the whole ``javax.json.JsonObject``, representing the whole PayoutMethod/Account, because we will need more information than just that flag and we want to avoid multiple useless API calls.


amihaiemil:
@rultor merge it please

Issue #550, opened 2020-09-15
Contributors.java:32-35: Implement and test...


0pdd:
The puzzle `548-3fa180bc` from #548 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/3d809b46ac30ac747f99ab75ec34c1916053286a/self-api/src/main/java/com/selfxdsd/api/Contributors.java#L32-L35

The puzzle was created by cristianpela on 11-Sep-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
@amihaiemil I'll work on this

amihaiemil:
paid with transaction id 4VB92593KY8088153

0pdd:
The puzzle `548-3fa180bc` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #556, opened 2020-09-19
#550. ProviderContributors implemented+tested.

criske:
PR for #550

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Contributors.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Contributors.java
@@ -29,10 +29,6 @@ import com.selfxdsd.api.storage.Paged;
  * @author Mihai Andronache (amihaiemil@gmail.com)
  * @version $Id$
  * @since 0.0.1
- * @todo #548:30min. Implement and test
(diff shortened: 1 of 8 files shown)


criske:
@amihaiemil I've noticing something odd. Why do we add a 0 hourly rate `Contract` in `ProjectContributors#register`?
https://github.com/self-xdsd/self-core/blob/a27f1cd4cafd99bb154194f916e4f3f764406ace/self-core-impl/src/main/java/com/selfxdsd/core/contributors/ProjectContributors.java#L130

amihaiemil:
@criske 
> Why do we add a 0 hourly rate Contract in ProjectContributors#register

We first register him as a Contributor and then we have to register a Contract with the Project, otherwise he will not be a project contributor. We just assume the default role to be ``DEV`` and default hourly rate to be ``0``.

We will change this, most likely. I want to have a minimum hourly rate, say ``$10/h`` in any project and any role.

amihaiemil:
@criske the PR looks good, but it's a little too big...

amihaiemil:
@rultor merge it please

PR #561, opened 2020-09-25
#560 Env constants class for env variables

### LINKED ISSUE #560: Constants Class for EnvVariables
Let's have class Env which will contain all the variables used by Self as String constants.

amihaiemil:
Fixes #560


amihaiemil:
@criske I made this constants class because we already have quite a few env variables and it's hard to keep track of them :D

Issue #564, opened 2020-09-26
StoredProjectManager.java:45-47: When reviewing the...


0pdd:
The puzzle `562-63248b5a` from #562 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/f5ca9431a732f117e579293b99c93e2e9c2b8561/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java#L45-L47

The puzzle was created by amihaiemil on 26-Sep-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
@amihaiemil I'll work on this

amihaiemil:
@criske sure, go ahead :D

amihaiemil:
paid with transaction ID 8LD813704J098315R

0pdd:
The puzzle `562-63248b5a` has disappeared from the source code, that's why I closed this issue. //cc @amihaiemil

PR #569, opened 2020-09-29
#564. StoredProjectManager#assignedTasks handle missing task deadline

criske:
- added test for this use case.
- added test for deadline reminder.
- ~~re-added "slash" at the end of `commentsUri` in GithubIssueComments ctor. Without it, some
GithubIssueComments tests fail (Uri#resolve used in those tests will eat the 'comments' uri path segment)~~

PR for #564

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
@@ -57,7 +57,7 @@ final class GithubIssueComments implements Comments {
         if(issueUriStr.endsWith("/")){
             slash = "";
         }
-        this.commentsUri = URI.create(issueUriStr + slash + "comments");
+        this.commentsUri = URI.create(issueUriStr + slash + "comments/");
         this.resources = resources;
     }
 
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
@@ -57,7 +57,7 @@ final class GithubIssueComments implements Comments {
         if(issueUriStr.endsWith("/")){
             slash = "";
         }
-        this.commentsUri = URI.create(issueUriStr + slash + "comments/");
+        this.commentsUri = URI.create(issueUriStr + slash + "comments");
         this.resources = resources;
     }
 
--- a/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
@@ -34,6 +34,7 @@ import java.math.BigDecimal;
 import java.time.LocalDateTime;
 import java.time.Period;
 import java.util.UUID;
+import java.util.function.Supplier;
 
 /**
  * A Project Manager stored in Self. Use this class when implementing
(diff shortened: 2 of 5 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
```diff
@@ -304,10 +341,23 @@ public void assignedTasks(final Event event) {
                         task.deadline().toLocalDate()
                     ).getDays();
                     final int left = Period.between(
-                        LocalDateTime.now().toLocalDate(),
+                        this.dateTimeSupplier.get().toLocalDate(),
                         task.deadline().toLocalDate()
                     ).getDays();
-                    if(left <= time/2) {
+                    if (left < 0) {
```
criske, line 347: @amihaiemil 
is this check in days too "relaxed"? or should we use a stricter check like `left` in seconds/minutes <=0?

self-core-impl/src/main/java/com/selfxdsd/core/GithubIssueComments.java
```diff
@@ -57,7 +57,7 @@
         if(issueUriStr.endsWith("/")){
             slash = "";
         }
-        this.commentsUri = URI.create(issueUriStr + slash + "comments");
+        this.commentsUri = URI.create(issueUriStr + slash + "comments/");
```
criske, line 60: @amihaiemil I've added `/` back, on my end some tests failed without it

amihaiemil: @criske trailing ``/`` is not ok, Github returns 404 because of it -- I found this bug and tested manually, but I was too lazy to write a test, sorry :D 

So please remove it and add a todo to fix the tests later.

self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
amihaiemil: @criske yeah, maybe we could take into account the hours/minutes as well. Can you leave a puzzle?

self-core-impl/src/test/resources/responses_en.properties
```diff
@@ -35,4 +35,5 @@ taskRegistered.comment=@%s I've just registered this ticket as a task and will a
 taskInvoiced.comment=@%s thank you for resolving this ticket. I've just added it to your active invoice. You can always\
                      check all your invoices and more on the [Contributor Dashboard](https://self-xdsd.com).
 taskDeadlineReminder.comment=@%s Don't forget to close this ticket before the deadline (%s). You are past the first half \
-                             of the allowed period.
\ No newline at end of file
+                             of the allowed period.
+taskDeadlineMissed.comment=@%s Looks like you've missed the task deadline (%s). You are now resigned from this task.
```
amihaiemil, line 39: @criske Can you add more text here. Say: 

```
taskDeadlineMissed.comment=@%s Looks like you've missed the task deadline (%s). You are now resigned from this task.\n\n\
                           Please stop working on it, you will not be paid. I will assign it to someone else soon.
```

amihaiemil (changes requested):
@criske see my comments :D

amihaiemil:
@criske now Travis complains because those tests are failing. Can you ignore them? :D

criske:
@amihaiemil Can I fix them in this pr? :D It should be a pretty short fix and are only 2 tests.

amihaiemil:
@criske sure

amihaiemil:
@rultor merge it please

PR #580, opened 2020-10-02
#578 register resignation, then resign

### LINKED ISSUE #578: Register resignation, then resign, when reviwing assigned tasks
When reviewing assigned tasks, if the deadline has passed, we should register the resignation and only after that resign. Currently it's the other way around and it could cause an exception: https://github.com/self-xdsd/self-core/blob/master/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java#L351

amihaiemil:
Fixes #578


amihaiemil:
@rultor merge it

PR #600, opened 2020-10-10
#587 When registering task, check for Issue closed

### LINKED ISSUE #587: Before registering a new task we should check if the issue is not closed.
When processing "register" event we should check if the issue is not closed first.

https://github.com/self-xdsd/self-core/blob/41e3831136a9c542851ff74aa18c444019fe62c2/self-core-impl/src/main/java/com/selfxdsd/core/managers/Register.java#L69

amihaiemil:
Fixes #587 

When registering an issue as a task, we should first check if the Issue is closed or not.
Added IssueIsClosed precondition Step and unit tests for it.


amihaiemil:
@rultor merge it

Issue #598, opened 2020-10-10
Wallet.java:36-40: A project's Wallet should have one or...


0pdd:
The puzzle `594-abba5903` from #594 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/5161988f24fc0062457b84299fb01495be09cefd/self-api/src/main/java/com/selfxdsd/api/Wallet.java#L36-L40

The puzzle was created by @amihaiemil on 10-Oct-20. 

Estimate: 60 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

charlesmike:
@0pdd thank you for reporting this. I'll assign someone to take care of it soon.

0pdd:
> @0pdd thank you for reporting this. I'll assign someone to take care of it soon.

  I see you're talking to me, but I can't reply since I'm not a chat bot.

charlesmike:
@criske this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2020-10-20T11:19:50.776717``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything - you can solve it only partially and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

amihaiemil:
paid with transaction id 7GP62353WR547894X

0pdd:
The puzzle `594-abba5903` has disappeared from the source code, that's why I closed this issue. //cc @charlesmike

charlesmike:
@criske thank you for resolving this ticket. I've just added it to your active invoice. You can always check all your invoices and more on the [Contributor Dashboard](https://self-xdsd.com).

PR #607, opened 2020-10-10
#598 PaymentMethods API and StoredPaymentMethod.

criske:
PR for #598
- left puzzle for PaymentMethods#ofWallet

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/projects/StoredPaymentMethod.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/projects/StoredPaymentMethod.java
@@ -51,18 +51,27 @@ public final class StoredPaymentMethod implements PaymentMethod {
      */
     private final Wallet wallet;
 
+
+    /**
+     * Is the payment method active or not.
+     */
+    private final boolean active;
+
     /**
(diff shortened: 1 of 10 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/projects/StoredPaymentMethod.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public Wallet wallet() {
+        return this.wallet;
+    }
+
+    @Override
+    public boolean active() {
+        return this.equals(this.storage.paymentMethods()
```
criske, line 80: @amihaiemil Should we provide this boolean in constructor instead, might be more efficient.

amihaiemil: @criske yes, definetely :D we do the same for wallets and PayoutMethods

amihaiemil (changes requested):
@criske looks nice. Just one comment, please provide the ``active`` flag in the constructor.

amihaiemil:
@rultor merge it please

Issue #603, opened 2020-10-10
UnassignTask.java:39-41: When unassigning the Task, we...


0pdd:
The puzzle `585-676e2a39` from #585 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/ca025e0e79ff1874809af4929010d9b514b3b32c/self-core-impl/src/main/java/com/selfxdsd/core/managers/UnassignTask.java#L39-L41

The puzzle was created by amihaiemil on 10-Oct-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

charlesmike:
@0pdd thank you for reporting this. I'll assign someone to take care of it soon.

0pdd:
> @0pdd thank you for reporting this. I'll assign someone to take care of it soon.

  I see you're talking to me, but I can't reply since I'm not a chat bot.

charlesmike:
@amihaiemil this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2020-10-20T15:19:51.092984``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything - you can solve it only partially and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

0pdd:
The puzzle `585-676e2a39` has disappeared from the source code, that's why I closed this issue. //cc @charlesmike

charlesmike:
@amihaiemil thank you for resolving this ticket. I've just added it to your active invoice. You can always check all your invoices and more on the [Contributor Dashboard](https://self-xdsd.com).

PR #608, opened 2020-10-11
#603 UnassinTask removes the Issue assignee as well

amihaiemil:
PR for #603 

When unasigning a Task we should also remove the assignee from the Issue.
Better deregister comment.
Updated test.


amihaiemil:
@rultor merge it please

PR #638, opened 2020-10-21
Repo Star for github #630

### LINKED ISSUE #630: Add Method Repo.star()
We should implement this method for GithubRepo, as explained here: https://developer.github.com/v3/activity/starring/#star-a-repository-for-the-authenticated-user

The star() method in GitlabRepo should just throw ``UnsupportedOperationException`` for now.

Since the URI does not respect the normal fluency, you will have to just build it entirely in the method (https://api.github.com/user/starred/:owned/:name). 

You can get the repo owner and name by parsing the URI encapsulated by GithubRepo and taking the last two path parts.

A unit tests using MkContainer should be enough.

ak2703:
Fixes #630

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Repo.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Repo.java
@@ -81,4 +81,10 @@ public interface Repo {
      * @return Webhooks.
      */
     Webhooks webhooks();
+
+    /**
+     * Star this repo.
+     * @return Stars
+     */
+    Stars star();
 }
--- a/self-api/src/main/java/com/selfxdsd/api/Stars.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Stars.java
@@ -1,0 +1,14 @@
+package com.selfxdsd.api;
+
+/**
+ * Repo stars.
+ */
+public interface Stars {
+
+    /**
+     * Star a repository for the authenticated user.
+     * @return True or false, whether the starring
+     *  was successful or not.
+     */
+    boolean star();
+}
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
@@ -1,0 +1,88 @@
+package com.selfxdsd.core;
+
+import com.selfxdsd.api.Stars;
+import com.selfxdsd.api.storage.Storage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.json.Json;
+import java.net.HttpURLConnection;
+import java.net.URI;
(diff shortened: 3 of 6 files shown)


self-api/src/main/java/com/selfxdsd/api/Repo.java
```diff
@@ -81,4 +81,10 @@
      * @return Webhooks.
      */
     Webhooks webhooks();
+
+    /**
+     * Star this repo.
+     * @return Stars
+     */
+    Stars star();
```
amihaiemil, line 89: @ak2703 Can you name this method ``stars()`` (plural)? That's the correct way, since it's the entry point of a Repo's Stars. 

Besides, as it is now, starring a repo will look weird: ``repo.star().star()``.

self-api/src/main/java/com/selfxdsd/api/Stars.java
```diff
@@ -0,0 +1,14 @@
...
+ * Repo stars.
+ */
+public interface Stars {
+
+    /**
+     * Star a repository for the authenticated user.
+     * @return True or false, whether the starring
+     *  was successful or not.
+     */
+    boolean star();
```
amihaiemil, line 13: @ak2703 Can you rename this method to ``add()``? Starring a repo should look like this: ``repo.stars().add()``.

amihaiemil (changes requested):
@ak2703 Just 2 cosmetic comments :)

self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
```diff
@@ -0,0 +1,88 @@
...
+            final URI starsUri,
+            final Storage storage
+    ) {
+        this.resources = resources;
+        this.starsUri = starsUri;
+        this.storage = storage;
+    }
+
+    @Override
+    public boolean star() {
```
amihaiemil, line 59: @ak2703 Please add a puzzle for the unit testing of this method. Normally, I would reject a PR without unit tests, but this is a simple usecase.

A "puzzle" is a ``todo`` and in this case it should like this:

```java
@todo #630:30min Please provide some unit tests for method star() here.
 We should start a mock server and catch the incoming request to assert
 that it is constructed correctly.
```

You can add it in the javadoc of the class. Please pay attention to its grammar: lines 2..n should start with a space (they should be aligned with the ``t`` instead of the ``@``).

amihaiemil (commented):
@ak2703 one more comment

amihaiemil:
@rultor merge it please

amihaiemil:
@ak2703 I've decided to merge this PR and I'll take care of the changes, since we're in a bit of a hurry. 

I've sent the payment via PayPal. Transaction ID is 1N03482294543235X.

Issue #632, opened 2020-10-18
GithubInvitation.java:40-41: After we have the method...


0pdd:
The puzzle `629-041ddd1b` from #629 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/87796b04c713eadae872281bb545bcba1dd47f1f/self-core-impl/src/main/java/com/selfxdsd/core/GithubInvitation.java#L40-L41

The puzzle was created by @amihaiemil on 18-Oct-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@charlesmike please pay attention to this new issue.

charlesmike:
@0pdd thank you for reporting this. I'll assign someone to take care of it soon.

0pdd:
> @0pdd thank you for reporting this. I'll assign someone to take care of it soon.

  I see you're talking to me, but I can't reply since I'm not a chat bot.

amihaiemil:
waiting for #630

charlesmike:
@criske this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2020-10-28T12:09:50.755590``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything - you can solve it only partially and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

criske:
@amihaiemil What will happen if PR for task is created before the deadline but merged after it?

amihaiemil:
@criske the PR has to be merged *before* the deadline. If the deadline passes, the Task is taken away and no payments will be made to the Contributor -- payments will be made only to the PR reviewer (that is treated as a different Task).

Maybe we could allow the PO or ARCH to reassign a Contributor even if they were resigned.

criske:
@amihaiemil But can't this be abused? The contributor submits the pr before the deadline, but the reviewer waits to pass the deadline and only they will be paid :D

amihaiemil:
@criske Well, yes. But Self is supposed to be a platform for teams that already work together and trust each other.

[...]

PR #644, opened 2020-10-26
#629 Added starring to GithubInvitation#accept + tested.

criske:
PR for #632

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GithubInvitation.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GithubInvitation.java
@@ -23,7 +23,6 @@
 package com.selfxdsd.core;
 
 import com.selfxdsd.api.Invitation;
-import com.selfxdsd.api.Provider;
 import com.selfxdsd.api.Repo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -65,7 +64,7 @@ final class GithubInvitation implements Invitation {
     /**
      * Parent Github.
      */
-    private final Provider github;
+    private final Github github;
 
     /**
      * Ctor.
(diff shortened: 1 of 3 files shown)


self-core-impl/src/main/java/com/selfxdsd/core/GithubInvitation.java
```diff
@@ -66,7 +65,7 @@
     /**
      * Parent Github.
      */
-    private final Github github;
+    private final Provider github;
```
amihaiemil, line 68: @criske This class represents a Github Invitation, why does this attribute have to be a ``Provider``?

amihaiemil (changes requested):
@criske Just 1 question

self-core-impl/src/main/java/com/selfxdsd/core/GithubInvitation.java
criske: @amihaiemil I've needed to verify in tests that `Stars#add` is called or not (a concrecte impl of `Provider` wouldn't let me as it can't be mocked). But I can change it back and check it with `MockJsonResources` if stars endpoint is called.

amihaiemil: @criske yes please, the MockJsonResources version would be better. I don't want to have workarounds in the code just because of the tests :D

self-core-impl/src/main/java/com/selfxdsd/core/GithubRepo.java
```diff
@@ -122,7 +122,11 @@ public Webhooks webhooks() {
     public Stars stars() {
         return new GithubStars(
             this.resources(),
-            this.repoUri(),
+            URI.create(this
+                .repoUri()
+                .toString()
+                .replace("repos", "user/starred")
+            ),
```
criske, line 129: @amihaiemil Here was a bug - was passed the repo url of  `/user/starred/:owner/:repo`

amihaiemil:
@criske thanks for fixing it!

amihaiemil:
@rultor merge it

PR #645, opened 2020-10-26
#640 do not elect contracts marked for removal

### LINKED ISSUE #640: If Contract is Marked For Removal, Do Not Elect
We should modify the election algorithm to not elect a Contract which is marked for deletion (Contract.markedForRemovel != null).

amihaiemil:
Fixes #640


amihaiemil:
@rultor merge it

PR #650, opened 2020-10-30
#649 Add PO Contract When Activating Repo.

### LINKED ISSUE #649: Add PO Contract When Activating Repo
When a user activates their Repo, we should also add them as a Contributor and register a PO Contract (hourly rate $25 by default).

criske:
PR for #649 

closes #649

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/BaseRepo.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/BaseRepo.java
@@ -148,6 +148,18 @@ abstract class BaseRepo implements Repo {
                 .replace("-", "")
         );
         wallets.activate(wallet);
+        final Contributors contributors = project.contributors();
+        contributors.register(
+            this.owner.username(),
+            this.owner.provider().name()
+        );
+        project.contracts().addContract(
+            project.repoFullName(),
+            this.owner.username(),
+            this.owner.provider().name(),
+            BigDecimal.valueOf(2500),
+            Contract.Roles.PO
+        );
         return project;
     }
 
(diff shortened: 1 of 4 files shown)


self-core-impl/src/test/java/com/selfxdsd/core/managers/StatusTestCase.java
```diff
@@ -131,7 +131,7 @@ public void repliesTaskNotRegisteredAndIssueClosed() {
                 + "@mihai this ticket is not registered as a task, "
                 + "therefore I'm not working on it at the moment.\n\n"
                 + "If you want me to take care of it, please reopen it."
-                + " However, reopening Issues is not encouraged --"
+                + " However, reopening Issues is not encouraged -- "
```
criske, line 134: @amihaiemil small fix here -- this test was failing.

self-core-impl/src/main/java/com/selfxdsd/core/BaseRepo.java
```diff
@@ -148,6 +148,18 @@ public Project activate() {
...
+            this.owner.username(),
+            this.owner.provider().name()
+        );
+        project.contracts().addContract(
+            project.repoFullName(),
+            this.owner.username(),
+            this.owner.provider().name(),
+            BigDecimal.valueOf(2500),
+            Contract.Roles.PO
+        );
```
amihaiemil, line 162: @criske I remember we once implemented a functionality, to automatically register the Contributor when adding a contract (if it's their first Contract in Self). Will not this work here? Maybe ``addContract(...)`` is enough. What do you think?

amihaiemil (changes requested):
@criske just one question

self-core-impl/src/main/java/com/selfxdsd/core/BaseRepo.java
criske: @amihaiemil Yep, you're right, just checked the storage impl :D

amihaiemil:
@rultor merge it

PR #655, opened 2020-10-31
#653 RegisterResignation implemented, tested and used in Resign

### LINKED ISSUE #653: Implement RegisterResignation Step
We need this intermediary step.
At the moment, step ``UnassignTask`` regsiter an ``ASKED`` Resignation.

However, we don't always have to register a resignation when unassigning a task: for exapmle, if the Task is deregistered by the PO, we should not register a resignation of the assignee, since it's not their decision to resign and they also did not miss any deadline.,

amihaiemil:
Fixes #653 

UnassignTask does not register the resignation anymore. 
Implemented and tested RegisterResignation.
Used RegisterResignation in the Resign flow.


amihaiemil:
@rultor merge it please

Issue #651, opened 2020-10-30
NPE in Pull_Request Event


amihaiemil:
Let's fix it.

```java
java.lang.NullPointerException: null
        at com.selfxdsd.core.GithubIssues.received(GithubIssues.java:123) ~[self-core-impl-0.0.30.jar!/:na]
        at com.selfxdsd.selfpm.Webhooks$1.issue(Webhooks.java:138) ~[classes!/:0.0.4-SNAPSHOT]
        at com.selfxdsd.core.managers.IssueAssigneeHasRoles.<init>(IssueAssigneeHasRoles.java:52) ~[self-core-impl-0.0.30.jar!/:na]
        at com.selfxdsd.core.managers.StoredProjectManager.newIssue(StoredProjectManager.java:228) ~[self-core-impl-0.0.30.jar!/:na]
        at com.selfxdsd.core.projects.StoredProject.resolve(StoredProject.java:172) ~[self-core-impl-0.0.30.jar!/:na]
        at com.selfxdsd.selfpm.Webhooks.github(Webhooks.java:106) ~[classes!/:0.0.4-SNAPSHOT]
```

Edit: will be completely fixed in self-xdsd/self-pm#54

charlesmike:
@amihaiemil thank you for reporting this. I'll assign someone to take care of it soon.

charlesmike:
@amihaiemil this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2020-11-09T11:55:26.734834``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it partially and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

charlesmike:
@amihaiemil Don't forget to close this ticket before the deadline (2020-11-09T11:55:26). You are past the first half of the allowed period.

charlesmike:
@amihaiemil thank you for resolving this ticket. I've just added it to your active invoice. You can always check all your invoices and more on the [Contributor Dashboard](https://self-xdsd.com).

PR #657, opened 2020-11-04
#651 changed GithubIssue.isPullRequest() logic

amihaiemil:
PR for #651

We needed to change this logic because the JSON that comes from the Webhook pull_request event does not have the ``pull_request`` child object.


amihaiemil:
@rultor merge it please

PR #661, opened 2020-11-04
#660 GithubIssue.repoFullName() from url attribute

### LINKED ISSUE #660: GithubIssue: Get RepoFullName from URL
At the moment we read the repoFullname from the ``repository_url`` attribute. However, the ``pull_request`` JSON that comes from a webhook event does not have this attribute.

We should instead get the repoFullName from the ``url`` attribute which is present everywhere.

amihaiemil:
Fixes #660


amihaiemil:
@rultor merge it please

PR #715, opened 2020-11-24
#714 100k for the Fake Wallet

### LINKED ISSUE #714: Register Only 100K for the Fake Wallet
At the moment we create fake wallets with 10M, which is way too much. 
We should register them with 100k, it should be more than enough for any user to get the grip on Self and decide if they want to use it or not.

amihaiemil:
Fixes #714


charlesmike:

Issue #303, opened 2020-07-18
Gitlab.java:36-37: Gitlab Invitations: implement and test...


0pdd:
The puzzle `300-8b838371` from #300 has to be resolved: 

https://github.com/self-xdsd/self-core/blob/69a45e5dafea387c476f8c9faed412aac342a736/self-core-impl/src/main/java/com/selfxdsd/core/Gitlab.java#L36-L37

The puzzle was created by cristianpela on 18-Jul-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@amihaiemil please pay attention to this new issue.

criske:
Just took a look over the api, and I don't think there is a way to list and accept invitations. You can accept only via email,

amihaiemil:
@criske yes, indeed, there is no way to accept invitations automatically.

But, the ``Add Member`` endpoint which we use in GitlabCollaborators.add(...) already adds the PM with developer permissions to the Project and there is no need to accept anything. 

We will implement these Invitations but make them do nothing, maybe just log something.

charlesmike:
Puzzle disappeared from the code, that's why I closed this ticket.

PR #749, opened 2020-12-01
#303 GitLab Invitations implemented and tested

amihaiemil:
PR for #303 

Implemented an empty GitlabRepoInvitations, because in the case of Gitlab there are no invitations to accept ever (when we invite the PM, they are automatically added as a Member).


charlesmike:

PR #766, opened 2020-12-07
#764 less amount at destination

### LINKED ISSUE #764: Application Fee Received in Contributor's Currency
At the moment we transfer all the money to the Connected Account (contributor's account) and specify an application fee which comes back to us.

However, this mechanism is not ok because the fee comes back to us in the currency of the Connected Account.
We should change the mechanism as explained [here](https://stripe.com/docs/connect/currencies#application-fees-for-destination-charges-and-converting-balances).

amihaiemil:
Fixes #764 

Before, we were sending the whole amount to the Contributor and ask Stripe to give us back our commission (application fee). The problem with this approach is that the application fee comes back to us in the currency of the Contributor's account.

Now, we are sending the Contributor only his revenue (Invoice.amount), which means we are keeping on to our commission directly (``Invoice.totalAmount minus Invoice.amount``)


charlesmike:

Issue #787, opened 2020-12-12
GitlabIssueLabels.java: Similarly to how GithubIssueL... 


charlesmike:
The puzzle ``723-c5aa0901`` originating from #723 has to be resolved:

https://github.com/self-xdsd/self-core/blob/331f4283db917c630338914b4f8a2a23289c4802/self-core-impl/src/main/java/com/selfxdsd/core/GitlabIssueLabels.java#L49-L53

"Similarly to how GithubIssueLabels.add(...) is implemented, we should modify the add(...) method here to first add the labels to the repository, so they are colored differently. At the moment, they will automatically be created by Gitlab at Repo level, but they will always be blue.".

It was created by amihaiemil on 2020-12-12T11:18:33Z. Estimation is ``30 minutes``.
If you have any technical questions, don't ask me, I won't be able to help. Open new issues instead.

charlesmike:
@amihaiemil this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2020-12-22T13:25:50.665993``.

Estimation here is ``30 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it partially and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

charlesmike:
Puzzle disappeared from the code, that's why I closed this ticket.

charlesmike:
@amihaiemil thank you for resolving this ticket. I've just added it to your active invoice. You can always check all your invoices and more on the [Contributor Dashboard](https://self-xdsd.com).

PR #789, opened 2020-12-12
#787 add labels to repo first

amihaiemil:
PR for #787 

Before adding labels to an Issue, we add them to the Repo first, so we can specify their color.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GitlabIssueLabels.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GitlabIssueLabels.java
@@ -46,11 +46,6 @@ import java.util.stream.Collectors;
  * @author Mihai Andronache (amihaiemil@gmail.com)
  * @version $Id$
  * @since 0.0.39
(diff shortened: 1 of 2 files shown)


charlesmike:

self-core-impl/src/main/java/com/selfxdsd/core/GitlabIssueLabels.java
```diff
@@ -94,9 +89,18 @@
...
+                this.uri.toString()
+                    .replaceAll("/issues/[0-9]+", "/labels")
+            ),
+            this.resources
+        );
+        repoLabels.add(names);
         final boolean added;
         final String labels = Arrays.stream(names)
             .collect(Collectors.joining(","));
+        System.out.println("Issue Labels: " + this.uri.toString());
```
fellahi-ali, line 103: @amihaiemil Is this intended or just forgotten, because it's already logged?

amihaiemil: @fellahi-ali it's forgotten there, I'll remove it now. Thanks!

PR #814, opened 2020-12-20
#811 Implemented Invoice.billedBy and billedTo

### LINKED ISSUE #811: Invoice.java: Add methods String::from() an...
The puzzle ``803-781023a5`` originating from #803 has to be resolved:

https://github.com/self-xdsd/self-core/blob/ed9cab39eb3c2c555720b464ca26f51548ab6cce/self-api/src/main/java/com/selfxdsd/api/Invoice.java#L13-L16

"Add methods String::from() and String::to(). StoredInvoice should accept these values in the constructor. If these values are null, they should be read from Project.billingInfo and Contributor.billingInfo. The Project and the Contributor are accessible via the Contract.".

It was created by @amihaiemil on 2020-12-20T13:22:33Z. Estimation is ``60 minutes``.
If you have any technical questions, don't ask me, I won't be able to help. Open new issues instead.

amihaiemil:
Fixes #811 

Implemented Invoice billedBy and billedTo.
If the two values are not set in the constructor, then they are read from the Contract.
The two values are written in the DB (thus can never change) when the invoice is paid.

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Invoice.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Invoice.java
@@ -10,10 +10,6 @@ import java.time.LocalDateTime;
  * @author criske
  * @version $Id$
  * @since 0.0.3
- * @todo #803:60min Add methods String::from() and String::to(). StoredInvoice
- *  should accept these values in the constructor. If these values are null,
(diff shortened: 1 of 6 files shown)


charlesmike:

PR #832, opened 2020-12-25
#830 BillingInfo.other is optional

### LINKED ISSUE #830: "Other" BillingInfo Attribute is optional
This attribute is optional, therefore we should check for null and replace with empty string.
/
Screenshot 2020-12-25 at 17 30 49

(The null firstName and lastName is solved in self-web)

amihaiemil:
Fixes #830 
BillingInfo.other is optional both when creating and when reading Stripe data.


charlesmike:

PR #851, opened 2021-01-01
#850 PM Commission calculated as 0-scale, rounded up BigDecimal (cents)

### LINKED ISSUE #850: StoredProjectManager.commission() should return 0 scale and half up
At the moment the commission of the PM is calculated as BigDecimal with scale 2, yet BigDecimal already represents the cents (this means, at the moment the PM's commission can be even more precise then cents thus causing rounding errors sometimes).

Let's fix it. It should be a precise BigDecimal (scale 0, we don't go lower than cents) with rounding mode Half_up.

amihaiemil:
Fixes #850 

Before, we were calculating the commission with a 2-digit precision, which was wrong because BigDecimal is already the lowest denomination (cents), meaning it was possible to have commissions with a precision bellow cents, thus causing rounding errors.

Now we are calculating with a 0 scale and we are rounding half up, so we will always have an integer number of  cents.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
@@ -209,12 +209,15 @@ public final class StoredProjectManager implements ProjectManager {
 
     @Override
     public BigDecimal commission(final BigDecimal value) {
-        return value
-            .multiply(BigDecimal
+        return value.multiply(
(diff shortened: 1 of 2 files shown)


charlesmike:

charlesmike:
@amihaiemil please review this Pull Request. Deadline (when it should be merged or closed) is ``2021-01-04T18:28:50.255919``.

charlesmike:

PR #920, opened 2021-01-19
Commits#latest is now nullable.

criske:
Otherwise will throw index out of bounds exception if there are no commits

--
no bounty needed

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Commits.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Commits.java
@@ -39,7 +39,7 @@ public interface Commits extends Iterable<Commit> {
 
     /**
      * Get the latest Commit (the newest one).
-     * @return Commit.
+     * @return Commit or null if there are no commits.
      */
     Commit latest();
 }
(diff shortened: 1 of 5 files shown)


charlesmike:

criske:
@amihaiemil Anyway I've tested most of the gitlab features against a real repo. Only thing I've missed testing is processing webhook responses.
I hope from now on there won't problems with gitlab on front end anymore :D

charlesmike:
@amihaiemil please review this Pull Request. Deadline (when it should be merged or closed) is ``2021-01-22T14:16:15.364739``.

amihaiemil:
@criske This won't be necessary. We use the Commits API only on ``PUSH`` events received from Github or GitLab. 
If we have a PUSH event, it's clear that we already have at least one commit in the Repo. 

And if we make Commit.latest() nullable, then we also need to change the client code (code calling Commit.latest()) and add ``if null else`` statements... 

I would rather leave it as it is with no changes.

charlesmike:

PR #931, opened 2021-01-26
#929 GitlabIssueLabels is array of String, not JsonObject

### LINKED ISSUE #929: GitlabIssueLabels.java: ClassCastException on iterator().
Reproduce: this happens when pm tries to register a task, after a "register" command.

https://github.com/self-xdsd/self-core/blob/eeb02883804d4ce6448e7f1d5f140f2c3a671a68/self-core-impl/src/main/java/com/selfxdsd/core/GitlabIssueLabels.java#L153-L161
```java
java.lang.ClassCastException: class org.glassfish.json.JsonStringImpl cannot be cast to class javax.json.JsonObject 

(org.glassfish.json.JsonStringImpl and javax.json.JsonObject are in unnamed module of loader 'app')
```

amihaiemil:
Fixes #929


charlesmike:

Issue #979, opened 2021-02-09
Let's Start Implementing Payments


amihaiemil:
Payments will be an Iterable of Payment and will belong to an Invoice.

A Stripe payment of an Invoice may fail for some reason and we should register it. An Invoice will have one or more payments and it will be considered paid when the first successful Payment occurs.

A Payment should have a transactionId, timestamp, value, status (FAILED, SUCCESSFUL) and reason ("authentication required", "not enough funds" etc).

charlesmike:
@amihaiemil this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-02-19T19:49:05.351252``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

charlesmike:
@amihaiemil please keep in mind that manual assignment of tickets is a bad practice and it is discouraged. Next time, please let me elect the assignee.

amihaiemil:
@criske FYI

charlesmike:
@amihaiemil Don't forget to close this ticket before the deadline (2021-02-19T19:49:05). You are past the first half of the allowed period.

charlesmike:
@amihaiemil Looks like you've missed the task deadline (2021-02-19T19:49:05). You are now resigned from this task.

Please stop working on it, you will not be paid. I will assign it to someone else soon.

charlesmike:
@fellahi-ali this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-03-01T19:57:55.849239``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

fellahi-ali:
[...]

PR #1024, opened 2021-03-02
#979 Add Payment & Payments interfaces

fellahi-ali:
PR for #979
- Add Payment & Payments interfaces.
- Add StoredPayment implementation.
- StoredPayment doesn't contain any logic it's just a DTO so I don't know if it should be tested and how?

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Invoice.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Invoice.java
@@ -15,7 +15,7 @@ import java.time.LocalDateTime;
  * @since 0.0.3
  *
  * @todo #979:30min Add payments() method that returns Invoice's Payments and
- * its implementation in StoredInvoice.
+ *  its implementation in StoredInvoice.
  */
 public interface Invoice {
 
--- a/self-api/src/main/java/com/selfxdsd/api/Invoice.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Invoice.java
@@ -13,6 +13,9 @@ import java.time.LocalDateTime;
  * @author criske
  * @version $Id$
  * @since 0.0.3
+ *
+ * @todo #979:30min Add payments() method that returns Invoice's Payments and
+ * its implementation in StoredInvoice.
  */
 public interface Invoice {
 
--- a/self-api/src/main/java/com/selfxdsd/api/Payment.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Payment.java
@@ -43,7 +43,7 @@ public interface Payment {
 
     /**
      * The transaction id of the payment.
-     * @return String or null if the payment has FAILED.
+     * @return String or empty if the payment has FAILED.
      */
     String transactionId();
 
--- a/self-api/src/main/java/com/selfxdsd/api/Payment.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Payment.java
@@ -37,7 +37,7 @@ public interface Payment {
 
     /**
      * The {@link Invoice} to which this Payment is made.
-     * @return an {@link Invoice}
+     * @return Invoice
      */
     Invoice invoice();
 
@@ -56,19 +56,19 @@ public interface Payment {
     /**
      * The value of the Payment.<br>
      * Which is usually the total amount of the Invoice.
-     * @return a BigDecimal or null if the payment fail.
+     * @return BigDecimal or null if the payment fail.
      */
     BigDecimal value();
 
     /**
      * The status of the Payment (SUCCESSFUL or FAILED).
-     * @return a String.
+     * @return String.
      */
     String status();
 
     /**
      * The description of the reason why the Payment fail.
-     * @return a String or null if the Payment done successfully.
+     * @return String or null if the Payment done successfully.
      */
     String failReason();
 
@@ -78,7 +78,7 @@ public interface Payment {
      */
     class Status {
         /**
-         * private ctor.
+         * Private ctor.
          */
         private Status(){};
 
--- a/self-api/src/main/java/com/selfxdsd/api/Payment.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Payment.java
@@ -42,6 +42,12 @@ public interface Payment {
     Invoice invoice();
 
     /**
+     * The invoice sent to {@link Contributor} in case of successful payment.
+     * @return PlatformInvoice or null if the payment failed.
+     */
+    PlatformInvoice platformInvoice();
+
+    /**
      * The transaction id of the payment.
      * @return String or empty if the payment has FAILED.
      */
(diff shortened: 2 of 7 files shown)


zoeself:
@fellahi-ali thank you for your Pull Request. I'll assign someone to review it soon.

zoeself:
@amihaiemil please review this Pull Request. Deadline (when it should be merged or closed) is ``2021-03-05T19:07:17.776299``.

self-api/src/main/java/com/selfxdsd/api/Invoice.java
```diff
@@ -13,6 +13,9 @@
  * @author criske
  * @version $Id$
  * @since 0.0.3
+ *
+ * @todo #979:30min Add payments() method that returns Invoice's Payments and
+ * its implementation in StoredInvoice.
```
amihaiemil, line 18: @fellahi-ali second and rest of sentences should always start with a space :D

self-api/src/main/java/com/selfxdsd/api/Payment.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * The status of the Payment (SUCCESSFUL or FAILED).
+     * @return String.
+     */
+    String status();
+
+    /**
+     * The description of the reason why the Payment fail.
+     * @return String or null if the Payment done successfully.
```
amihaiemil, line 71: @fellahi-ali ``String or empty string if done successfully`` - we don't need to return null in this case :D

self-core-impl/src/main/java/com/selfxdsd/core/contracts/invoices/StoredPayment.java
```diff
@@ -0,0 +1,137 @@
...
+    @Override
+    public String failReason() {
+        return this.failReason;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(
+            this.invoice.invoiceId(),
+            this.paymentTime()
```
amihaiemil, line 128: @fellahi-ali paymentTime and transactionId should be enough, no need for the invoice's ID.

self-core-impl/src/main/java/com/selfxdsd/core/contracts/invoices/StoredPayment.java
```diff
@@ -0,0 +1,137 @@
...
+        );
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        return this == obj || (obj instanceof StoredPayment
+            && this.hashCode() == obj.hashCode());
```
amihaiemil, line 135: @fellahi-ali This method doesn't look right. The relation between equals and hashcode is:

1. If 2 objects are equal, then the hashcode must be identical.
BUT
2. If the hashcode is identical, it doesn't necessarily mean that the objects are equal. 

Maybe use the IDE to generate equals and hashcode based on paymentTime and transactionId?

self-api/src/main/java/com/selfxdsd/api/Wallet.java
```diff
@@ -63,6 +63,9 @@ default BigDecimal debt() {
      * Pay an invoice.
      * @param invoice The Invoice to be paid.
      * @return Wallet having cash deducted with Invoice amount.
+     *
+     * @todo #979:60min After Invoice.payments() is implemented use it to
+     *  register all payments made for an invoice failed and successful one.
      */
```
amihaiemil, line 69: @fellahi-ali You can add the return Payment todo here as well ;D

self-core-impl/src/main/java/com/selfxdsd/core/contracts/invoices/StoredPayment.java
```diff
@@ -0,0 +1,137 @@
...
+import java.time.LocalDateTime;
+import java.util.Objects;
+
+/**
+ * A Payment stored in Self.
+ * @author Ali FELLAHI (fellahi.ali@gmail.com)
+ * @version $Id$
+ * @since 0.0.67
+ */
+public final class StoredPayment implements Payment {
```
amihaiemil, line 38: @fellahi-ali These classes should also be tested, yes. In this case, it is a simple DTO, but in other cases you see these ``Stored*`` classes are encapsulating the ``Storage`` and using it to read data directly from the storage.

But you can write a todo for this unit testing.

self-api/src/main/java/com/selfxdsd/api/Payment.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * A Payment of an {@link Invoice}.
+ * <br>
+ * It represent a Payment that could be Successful or Failed.
+ * @author Ali FELLAHI (fellahi.ali@gmail.com)
+ * @version $Id$
+ * @since 0.0.67
+ */
+public interface Payment {
```
amihaiemil, line 36: @fellahi-ali Can you also add a todo for adding method ``PlatformInvoice :: platformInvoice()`` to this interface? I forgot about mentioning this method and it's really important to have it. 

If a Payment is successful (and done with the real wallet), it should have a PlatformInvoice attached to it, which is the invoice from Self XDSD to the Contributor, for the commission.

amihaiemil (changes requested):
@fellahi-ali looks good, just a few comments :D

self-api/src/main/java/com/selfxdsd/api/Invoice.java
fellahi-ali: ah, sorry, I know that but don't know how I missed it :(

self-core-impl/src/main/java/com/selfxdsd/core/contracts/invoices/StoredPayment.java
fellahi-ali: @amihaiemil I didn't use transactionId because it could be null if the Payment failed?

amihaiemil: @fellahi-ali right, but we need to use it. Let's just use it and make sure it will be empty string, not null.

fellahi-ali: @amihaiemil my concern here is that if we remove the `invoiceId` from hashCode() and equals() than all failed transactions (i.e: with empty `transactionId`) with same `paymentTime`  would be equal.

amihaiemil: @fellahi-ali right. Well, the paymentTime will be a timestamp so very kittle chance to have 2 Payments performes at the same millisecond.

But you're right, let's leave it there.

fellahi-ali:
@amihaiemil CR fixes pushed :)

amihaiemil:
@fellahi-ali thanks. I removed the todo from the InMemory class -- we will not continue implementing the InMemory classes, we don't need them anymore. They were just useful in the beginning, until the project caught its shape :D 

Now we should remove them, but we are not removing them because they are still used in some unit tests.

zoeself:

PR #1051, opened 2021-03-10
#1049 Better messages for WalletPaymentException

amihaiemil:
Revised messages of this exception because they will be saves as Payment failureReason and displayed to both the PO and the Contributor.


zoeself:

Issue #984, opened 2021-02-12
BitbucketOrganizationRepos.java: Continue implementing this cl... 


charlesmike:
The puzzle ``1962974179`` originating from #978 has to be resolved:

https://github.com/self-xdsd/self-core/blob/504a7a5402c38e0c9fd6d1a6d290d93be1c144b2/self-core-impl/src/main/java/com/selfxdsd/core/BitbucketOrganizationRepos.java#L19-L19

"Continue implementing this class and writing tests for it.".

The puzzle was created by @amihaiemil at ``2021-02-12 12:26:26 +0200``. 
Estimation is ``60 minutes``.
If you have any technical questions, don't ask me, I won't be able to help. Open new issues instead.

charlesmike:
@amihaiemil this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-02-22T12:30:24.639826``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

charlesmike:
@amihaiemil Don't forget to close this ticket before the deadline (2021-02-22T12:30:24). You are past the first half of the allowed period.

charlesmike:
@amihaiemil Looks like you've missed the task deadline (2021-02-22T12:30:24). You are now resigned from this task.

Please stop working on it, you will not be paid. I will assign it to someone else soon.

charlesmike:
@fellahi-ali this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-03-04T16:37:35.743333``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

charlesmike:
[...]

PR #1053, opened 2021-03-10
#984 BitbucketOrganizationRepos impl & test

fellahi-ali:
PR for #984 
This impl assumes that  `BitbucketOrganizationRepos` has the correct repos URI which is not based on the following code:  
https://github.com/self-xdsd/self-core/blob/36dbcb49c816783d5c692f5ef004812a70240ccf/self-core-impl/src/main/java/com/selfxdsd/core/BitbucketOrganization.java#L66-L73

because links in the json response have `!` prefix as in 
```
"repositories": {
                    "href": "https://bitbucket.org/!api/2.0/repositories/alilosoft"
               }
``` 
and as discussed earlier with @criske those cause an authentication problem.
I will file an issue about it + other issues I found when working on this task :)


zoeself:

amihaiemil:
@fellahi-ali thanks, looks good!

PR #1055, opened 2021-03-11
#1013 Start implementing BitbucketRepo::issues()

fellahi-ali:
PR for #1013
I think this PR is relatively big :( , because I found myself required to fill some gaps to keep the code consistent :)

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/BitbucketIssue.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/BitbucketIssue.java
@@ -1,0 +1,179 @@
+/**
+ * Copyright (c) 2020-2021, Self XDSD Contributors
+ * All rights reserved.
+ * <p>
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"),
+ * to read the Software only. Permission is hereby NOT GRANTED to use, copy,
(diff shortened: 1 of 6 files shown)


zoeself:
@fellahi-ali thank you for your Pull Request. I'll assign someone to review it soon.

zoeself:

amihaiemil:
@fellahi-ali looks good! 

Actually, I don't know how Bitbucket Issues will work. From what I see, there are no Issues in BitBucket... instead, you can have issues/tickets in BitBucket if you integrate Jira. (cc: @criske). 

So I'm not sure how BitBucketIssues will be implemented. But anyway, it's a good start, thanks!

criske:
@amihaiemil @fellahi-ali Jira issues not supported by Bitbucket API, at least for now. PO must activate the local issue tracker from "Repository Settings"  in order that Issues api to work.

issues


Otherwise we will get.
```json
{
    "type": "error",
    "error": {
        "message": "Repository has no issue tracker."
    }
}
```
Maybe we should throw ISE, when get this message instead of returning null on `Issues#getById`.
Also the exception message may instruct user on how to activate the issue tracker and warn that Jira is not supported.

fellahi-ali:
Thanks @criske for insights, I was actually struggling to use the issues API, but couldn't, I  only used/trusted the available doc to impl this task, and I don't know if we could go further?

Issue #1071, opened 2021-04-01
Implement The "commands" Command


amihaiemil:
We should implement the ``commands`` command. When the user says ``@zoeself commands``, the PM should reply with the commands which it can support, their keywords and explanation.

zoeself:
@criske this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-04-11T15:10:55.613900``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

zoeself:
@amihaiemil please keep in mind that manual assignment of tickets is a bad practice and it is discouraged. Next time, please let me elect the assignee.

amihaiemil:
@criske do you still remember this area of the code? :D 

If yes, please start on it when you have some time. Don't forget about todos, maybe 2h is not enough to implement everything.

criske:
@amihaiemil yep, it should be pretty straight forward for the command itself.

zoeself:
@criske thank you for resolving this ticket. I've just added it to your active invoice. You can always check all your invoices and more on the [Contributor Dashboard](https://self-xdsd.com).

PR #1072, opened 2021-04-02
#1071 `Commands` implemented + Language linked responses.

criske:
This PR also introduces linked language responses. This means that a value for a response key could be a link
to a separate file that contains the actual reply. The file, depending on used link protocol, could be fetched locally (`file://`, `classpath:`) or remotely (`http(s)://`).

By doing this, larger and/or marked up replies can be easly maintainted in separate files rather than having them clobbered in the `responses.properites`.

PR for #1071

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Language.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Language.java
@@ -25,9 +25,7 @@ package com.selfxdsd.api;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.net.MalformedURLException;
 import java.net.URL;
-import java.net.URLConnection;
 import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -38,21 +36,18 @@ import java.util.stream.Collectors;
  * @version $Id$
  * @since 0.0.8
  * @checkstyle ReturnCount (100 lines)
- * @todo #1071:60min Continue writing unit tests
- *  for {@link Language#tryReplyFromLink(String)} in order to ensure complete
- *  test coverage.
  */
 public abstract class Language {
 
     /**
      * Commands that the agent can understand, in a given language.
      */
-    private final Properties commands;
+    private final Properties commands = new Properties();
 
     /**
      * Responses that the agent can give, in a given language.
      */
-    private final Properties responses;
+    private final Properties responses = new Properties();
 
     /**
      * Constructor. These two files should be in self-pm, so we don't have
@@ -65,7 +60,6 @@ public abstract class Language {
         final String commandsFileName,
         final String responsesFileName
     ) {
-        this(new Properties(), new Properties());
         try {
             this.commands.load(
                 this.getClass().getClassLoader()
@@ -81,20 +75,6 @@ public abstract class Language {
                 ex
             );
         }
-    }
-
-    /**
-     * Generic constructor from existing command and responses properties.
-     * Should be used only in tests.
-     * @param commands Commands properties.
-     * @param responses Responses properties.
-     */
-    protected Language(
-        final Properties commands,
-        final Properties responses
-    ){
-        this.commands = commands;
-        this.responses = responses;
     }
 
     /**
@@ -128,8 +108,7 @@ public abstract class Language {
      * @return String reply or null if nothing is found.
      */
     public final String reply(final String key) {
-        final String reply = this.responses.getProperty(key);
-        return tryReplyFromLink(reply);
+        return followPossibleLink(this.responses.getProperty(key));
     }
 
     /**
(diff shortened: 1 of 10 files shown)


zoeself:

self-core-impl/src/test/java/com/selfxdsd/core/projects/LanguageLinkedResponsesTestCase.java
```diff
@@ -0,0 +1,113 @@
...
+     * @checkstyle JavadocMethod (40 lines).
+     */
+    @Test
+    public void shouldReadReplyFromHttpLink() {
+        final Properties commands = new Properties();
+        commands.putIfAbsent(
+            "commands.comment",
+            "https://raw.githubusercontent.com/self-xdsd/self-docs/"
+                + "master/projectmanager.md"
+        );
```
criske, line 101: @amihaiemil It will be nice is we could extract the Commands part into a separate file and then having in the responses.properties
commands.comment = https://raw.githubusercontent.com/self-xdsd/self-docs/master/commands.md
This way we will have a centralized `commands` documentation used by both self-docs and self-pms.

zoeself:

self-api/src/main/java/com/selfxdsd/api/Language.java
```diff
@@ -56,6 +65,7 @@ public Language(
         final String commandsFileName,
         final String responsesFileName
     ) {
+        this(new Properties(), new Properties());
```
amihaiemil, line 68: @criske Why is this needed? :D 

I would prefer to leave the constructor as it was initially.

self-api/src/main/java/com/selfxdsd/api/Language.java
```diff
@@ -104,7 +128,95 @@ public final String categorize(final String command) {
...
+     *
+     * If something goes wrong like IO exception, bad request etc...,
+     * it will return null.
+     *
+     * @param linkedReply Reply as link.
+     * @return Actual reply fetched from link or null if something goes wrong.
+     * @checkstyle CyclomaticComplexity (60 lines).
+     * @checkstyle BooleanExpressionComplexity (20 lines).
+     */
+    private String tryReplyFromLink(final String linkedReply){
```
amihaiemil, line 173: @criske Let's name this method ``followPossibleLink``. 

Also, I like the idea of linked files for bigger responses but this method is waaay over-engineered. The ``classpath://`` option is enough, please remove the rest of the logic. 

If we will ever read language resources from a webservice, we will need some serious architectural changes.

self-api/src/main/java/com/selfxdsd/api/Language.java
```diff
@@ -104,7 +128,95 @@ public final String categorize(final String command) {
      * @return String reply or null if nothing is found.
      */
     public final String reply(final String key) {
-        return this.responses.getProperty(key);
+        final String reply = this.responses.getProperty(key);
+        return tryReplyFromLink(reply);
```
amihaiemil, line 132: @criske just ``return followPossibleLink(this.responses.getProperty(key));`` should be enough.

self-core-impl/src/test/resources/replies/commands.comment_en.md
```diff
@@ -0,0 +1,5 @@
+* ``Hello`` -- any comment addressed to the PM containing the word **"hello"**.
+* ``Status`` -- any comment addressed to the PM, containing the word **"status"**. The PM will respond with the Task's status, assignee, deadline etc.
+* ``Resign`` -- a task's assignee can always resign from the task, if they cannot or simply do not want to solve it. The comment should be addressed to the PM and contain the word **"resign"**. Only the task's assignee can give this command.
+* ``Deregister`` -- a task can be deregistered (taken out of scope) with a comment containing the word **"deregister"**. Only a Contributor with role ``PO`` or ``ARCH`` can give this command. When a task is deregistered, the task's assignee is automatically resigned.
+* ``Register`` -- an Issue or PR can be registered as a Task with a comment containing the word **"register"**. Any Contributor can give this command.
```
amihaiemil, line 5: @criske Can you also start this message with: ``Hi @%s! Here are the commands which I understand: ``, where %s should be replaced with the author's username. 

Also, at Register and Deegister, we should specify ``add`` and ``remove`` too. If I remember well, also the Resign command has more options.

If you look into ``commands_en.properties``, some commands have more than one entry, which means the PM will respond with that command for all those entries.

amihaiemil (changes requested):
@criske looks nice overall, but I have some comments :D

amihaiemil:
@criske one more thing: I don't see the ``Commands`` class being used anywhere. Did you forget to add it in the Conversation composition? :D

amihaiemil:
@criske inca o chestie, mi-am amintit acum. Fisierele astea de proprietati si raspunsuri in core sunt resurse de Test. Ele trebuie copiate in self-pm ca sa functioneze. Deci mecanismu asta cu classpath:// ar trebui sa functioneze si acolo :D

self-api/src/main/java/com/selfxdsd/api/Language.java
criske: >The classpath:// option is enough, please remove the rest of the logic.

@amihaiemil  I agree the `file://` should be removed, classpath is enough for local files.

> If we will ever read language resources from a webservice, we will need some serious architectural changes.

@amihaiemil For this particular case, https://github.com/self-xdsd/self-docs is the webservice.
We only need to extract "Commands" listing part from https://raw.githubusercontent.com/self-xdsd/self-docs/master/projectmanager.md into a separate file, `commands.md` for ex. I think Jekyll supports including other files.
(something like `{%commands.md%}` in `projectmanager.md`)

Now on `self-pm` responses.properties we could have two entries.
```properties
commands.comment.header=Hi @%s! Here are the commands which I understand: 
commands.comment.body=https://raw.githubusercontent.com/self-xdsd/self-docs/master/commands.md
```
And then in `Commands` Conversation we merge the header with body to `SendReply`.
This way any changes on self-docs' `commands.md` will be reflected on the `commands` reply.

But until we have that refactoring on self-docs we could still use the `classpath` aproach. What do you think ? :D

self-api/src/main/java/com/selfxdsd/api/Language.java
criske: @amihaiemil Needed to test `file://` protocol, that works only with absolute paths (I couldn't make it to work otherwise).
Just adding a file full path in `responses.properties` will break the ci though.

criske:
> @criske one more thing: I don't see the `Commands` class being used anywhere. Did you forget to add it in the Conversation composition? :D

I forgot about that, sorry :D

self-api/src/main/java/com/selfxdsd/api/Language.java
amihaiemil: @criske I see your point but I don't agree. Self-Docs is just a static site for documentation, I don't want to mix them up together. :D 

Someone will surely end up modifying self-docs without knowing of this mechanism and break it. Please only use the classpath option.

amihaiemil: @criske If you want to make an integration and make sure everything is up to date, then we should generate the documentation from the code, not the other way around :D 

Maybe we can think about this later

amihaiemil:
@criske nice, thanks!

amihaiemil:
@rultor merge

Issue #1084, opened 2021-04-08
StripeWallet.java: Take in consideration the con... 


zoeself:
The puzzle ``68729944`` originating from #1079 has to be resolved:

https://github.com/self-xdsd/self-core/blob/84ae5e82793c236c6bfbee1dd8f8406eee495f9a/self-core-impl/src/main/java/com/selfxdsd/core/projects/StripeWallet.java#L154-L156

"Take in consideration the contributor's commission when calculating the VAT and also deduct this commission from the sum which the contributor will receive, together with the VAT.".

The puzzle was created by amihaiemil at ``2021-04-08 19:44:13 +0300``. 
Estimation is ``60 minutes``.
If you have any technical questions, don't ask me, I won't be able to help. Open new issues instead.

zoeself:
@amihaiemil this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-04-18T17:01:09.309911``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

zoeself:
Puzzle disappeared from the code, that's why I closed this ticket.

zoeself:
@amihaiemil thank you for resolving this ticket. I've just added it to your active invoice. You can always check all your invoices and more on the [Contributor Dashboard](https://self-xdsd.com).

PR #1085, opened 2021-04-11
#1084 account for contributor commission when paying

amihaiemil:
PR for #1084 

Contributor commission is deducted from the Contributor's gross earnings;
Contributor commission is taken into account when calculating the VAT;
Added the reverse charge (negate the VAT) for EU contributors which are paying VAT;
If reverse charge applies, VAT is NOT deducted from the Contributor's earnings;
Left puzzle to update the PlatformInvoice's PDF.


zoeself:

PR #1088, opened 2021-04-11
#1083 StoredInvoice caches its InvoicedTasks

amihaiemil:
PR for #1083 

Easier solution: StoredInvoice has a Supplier of InvoicedTasks which reads them from the Storage only once and caches the result;
Added unit test;


zoeself:

PR #1106, opened 2021-04-24
#1104 treat UK as outside SEPA too

### LINKED ISSUE #1104: Treat the UK as non-SEPA
At the moment we treat UK as a SEPA country and Stripe seems to work fine. However, after a discussion with Stripe Support, we decided it's better to treat is a non-SEPA.

amihaiemil:
Fixes #1104


zoeself:

PR #1122, opened 2021-05-02
#938 Implementing TaskIssue

### LINKED ISSUE #938: Implement class TaskIssue (implements Issue)
We need to have an Issue implementation based on Task. 

Later, we will use ``E-Tags`` to read Issues from Github/GitLab API and the response will be 304 NOT MODIFIED (if there were no modifications) with an empty body. This will help us because these calls will not cound in the Rate Limit. 

For this scenario, we need class TaskIssue, to return the Issue based on the existing Task.

amihaiemil:
Fixes #938 

Implemented TaskIssue and added unit tests.
Left puzzles to continue work.


zoeself:

zoeself:

Issue #1123, opened 2021-05-02
GithubIssues.java: Fetch the JSON Issue only whe... 


zoeself:
The puzzle ``-1072508911`` originating from #938 has to be resolved:

https://github.com/self-xdsd/self-core/blob/83f672bcd8287c5da95d437ef89bd6bc57b88e91/self-core-impl/src/main/java/com/selfxdsd/core/GithubIssues.java#L47-L49

"Fetch the JSON Issue only when Issue.json() is called (use a caching Supplier inside GithubIssue to make sure the JSON is fetched only once).".

The puzzle was created by amihaiemil at ``2021-05-02 11:09:11 +0300``. 
Estimation is ``60 minutes``.
If you have any technical questions, don't ask me, I won't be able to help. Open new issues instead.

zoeself:
@amihaiemil this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-05-12T11:32:25.803997``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

zoeself:
Puzzle disappeared from the code, that's why I closed this ticket.

zoeself:
@amihaiemil thank you for resolving this ticket. I've just added it to your active invoice. You can always check all your invoices and more on the [Contributor Dashboard](https://self-xdsd.com).

PR #1126, opened 2021-05-02
#1123 GithubIssue works with Supplier of Json

amihaiemil:
PR for #1123 

We don't read the JSON issue directly when calling GithubIssues.getById(); instead, the JSON is read and cached into an encapsulated Supplier (inside GithubIssue) only when needed the first time.

Updated unit tests.


zoeself:

Issue #1135, opened 2021-05-03
JsonResources General Caching PoC


amihaiemil:
Proof of concept for general caching of json resources. Let's do it with clean Java SE, no dependencies (maybe define a JsonStorage which will save the JSONs in the DB)

@criske make a PoC PR and we will integrate it later if we need this caching in more places :D

zoeself:
@criske this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-05-13T14:36:07.688620``.

Estimation here is ``90 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

amihaiemil:
@zoeself remove

zoeself:
> @zoeself remove

@amihaiemil ok, I've removed this task from scope. I'm not managing it anymore.

amihaiemil:
@zoeself add

zoeself:
> @zoeself add

@amihaiemil I've just registered this ticket as a task and will assign it to someone soon. Thanks!

zoeself:
@criske this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-05-14T09:12:26.946269``.

Estimation here is ``180 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

zoeself:
@criske thank you for resolving this ticket. I've just added it to your active invoice. You can always check all your invoices and more on the [Contributor Dashboard](https://self-xdsd.com).

PR #1136, opened 2021-05-03
#1135. CachingJsonResources and JsonStorage implemented and tested.

criske:
pr for #1135

`CachingJsonResources` and `JsonStorage` implemented and unit tested.
Also tested against a real Github account over network.
In order to work in a decoupled fashion from concrete JsonResources implementations,
response headers are exposed from Resource.

Hooking up:
```java
new Github(user, storage, new CachingJsonResources(new JsonResources.JdkHttp()));
```
`JsonStorage` is not needed at the moment. `CachingJsonResources` is working with `JsonStorage.InMemory` just fine.
In the future, when we have a real caching, either with current db or other mechanisms, the setup would look like
`new CachingJsonResources(new JsonResources.JdkHttp(), betterJsonStorage)`

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
@@ -122,7 +122,7 @@ public final class CachingJsonResources implements JsonResources {
 
     @Override
     public Resource put(final URI uri, final JsonValue body) {
-        return this.delegate.patch(uri, body);
+        return this.delegate.put(uri, body);
     }
 
     @Override
--- a/self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
@@ -105,7 +105,7 @@ public final class CachingJsonResources implements JsonResources {
     @Override
     public Resource post(final URI uri,
                          final JsonValue body) {
-        return this.delegate.get(uri);
+        return this.delegate.post(uri, body);
     }
 
     @Override
--- a/self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
@@ -1,0 +1,241 @@
+/**
+ * Copyright (c) 2020-2021, Self XDSD Contributors
+ * All rights reserved.
+ * <p>
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"),
+ * to read the Software only. Permission is hereby NOT GRANTED to use, copy,
+ * modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software.
+ * <p>
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.selfxdsd.core;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.json.Json;
+import javax.json.JsonValue;
+import java.io.StringReader;
+import java.net.HttpURLConnection;
+import java.net.URI;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+/**
+ * Cacheable JSON Resources used by Provider.
+ * @author criske
+ * @version $Id$
+ * @since 0.0.79
+ */
+public final class CachingJsonResources implements JsonResources {
+
+    /**
+     * Logger.
+     */
+    private static final Logger LOG = LoggerFactory.getLogger(
+        CachingJsonResources.class
+    );
+
+
+    /**
+     * JsonResources delegate.
+     */
+    private final JsonResources delegate;
+
+
+    /**
+     * Json storage.
+     */
+    private final JsonStorage jsonStorage;
+
+    /**
+     * Ctor.
+     * @param delegate JsonResources delegate.
+     * @param jsonStorage JSON storage.
+     */
+    public CachingJsonResources(final JsonResources delegate,
+                                final JsonStorage jsonStorage) {
+        this.delegate = delegate;
+        this.jsonStorage = jsonStorage;
+    }
+
+    /**
+     * Ctor.
+     * @param delegate JsonResources delegate.
+     */
+    public CachingJsonResources(final JsonResources delegate) {
+        this(delegate, new JsonStorage.InMemory());
+    }
+
+
+    @Override
+    public JsonResources authenticated(final AccessToken accessToken) {
+        return new CachingJsonResources(
+            this.delegate.authenticated(accessToken),
+            this.jsonStorage
+        );
+    }
+
+    @Override
+    public Resource get(final URI uri) {
+        return this.tryGetFromCache(uri, Collections::emptyMap);
+    }
+
+    @Override
+    public Resource get(final URI uri,
+                        final Supplier<Map<String, List<String>>> headers) {
+        return this.tryGetFromCache(uri, headers);
+    }
+
+    @Override
+    public Resource post(final URI uri,
+                         final JsonValue body) {
+        return this.delegate.get(uri);
+    }
+
+    @Override
+    public Resource post(final URI uri,
+                         final Supplier<Map<String, List<String>>> headers,
+                         final JsonValue body) {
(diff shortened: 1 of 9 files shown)


zoeself:

zoeself:

amihaiemil:
@criske looks really interesting, thanks! I will have a look tomorrow :D

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
```diff
@@ -0,0 +1,241 @@
...
+import java.util.Map;
+import java.util.function.Supplier;
+
+/**
+ * Cacheable JSON Resources used by Provider.
+ * @author criske
+ * @version $Id$
+ * @since 0.0.79
+ */
+public final class CachingJsonResources implements JsonResources {
```
amihaiemil, line 45: @criske I think a better name is ``ConditionalJsonResources``. Apparently, the term for this E-Tag "caching" is "conditional requests". I've seen this in many places :D

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
```diff
@@ -0,0 +1,241 @@
...
+        this.jsonStorage = jsonStorage;
+    }
+
+    /**
+     * Ctor.
+     * @param delegate JsonResources delegate.
+     */
+    public CachingJsonResources(final JsonResources delegate) {
+        this(delegate, new JsonStorage.InMemory());
+    }
```
amihaiemil, line 83: @criske Can you move this constructor on top? The cascading constructors should always be the first ones.

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
```diff
@@ -0,0 +1,241 @@
...
+    public JsonResources authenticated(final AccessToken accessToken) {
+        return new CachingJsonResources(
+            this.delegate.authenticated(accessToken),
+            this.jsonStorage
+        );
+    }
+
+    @Override
+    public Resource get(final URI uri) {
+        return this.tryGetFromCache(uri, Collections::emptyMap);
```
amihaiemil, line 96: @criske Can we call the overloaded method instead? ``this.get(uri, () -> new HashMap<>());``. I think it's more straight forward.

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
```diff
@@ -0,0 +1,241 @@
...
+        );
+    }
+
+    /**
+     * Append "If-None-Match" header to current headers.
+     * @param headers Headers.
+     * @param etag Etag.
+     * @return Updated headers.
+     */
+    private Supplier<Map<String, List<String>>> updateHeaders(
```
amihaiemil, line 230: @criske Can you name this method simply ``ifNoneMatch``?

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
```diff
@@ -0,0 +1,241 @@
...
+     */
+    private Resource tryGetFromCache(
+        final URI uri,
+        final Supplier<Map<String, List<String>>> headers
+    ) {
+        final Resource resource;
+        String cachedEtag = this.jsonStorage.getEtag(uri);
+        if (cachedEtag != null) {
+            final Supplier<Map<String, List<String>>> updatedHeaders =
+                this.updateHeaders(headers, cachedEtag);
```
amihaiemil, line 148: @criske This variable is declared and only used once. Can we inline the method call instead?

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    /**
+     * Try to get Resource from json storage cache by URI, otherwise fetch
+     * from remote and then store the Resource.
+     * @param uri URI.
+     * @param headers Current Headers.
+     * @return Cached or remote Resource.
+     */
+    private Resource tryGetFromCache(
```
amihaiemil, line 140: @criske I would name this method ``conditionalGet``

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
```diff
@@ -0,0 +1,241 @@
...
+        if (cachedEtag != null) {
+            final Supplier<Map<String, List<String>>> updatedHeaders =
+                this.updateHeaders(headers, cachedEtag);
+            final Resource remoteResource = this.delegate
+                .get(uri, updatedHeaders);
+            final int status = remoteResource.statusCode();
+
+            if (status == HttpURLConnection.HTTP_NOT_MODIFIED) {
+                final String cachedBody = this.jsonStorage.getResourceBody(uri);
+                if (cachedBody != null) {
```
amihaiemil, line 155: @criske Is this if/else really necessary? If the status code is NOT_MODIFIED, then we should always find the body in the storage, or?

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
```diff
@@ -0,0 +1,241 @@
...
+     * @param uri URI.
+     * @param headers Current Headers.
+     * @return Cached or remote Resource.
+     */
+    private Resource tryGetFromCache(
+        final URI uri,
+        final Supplier<Map<String, List<String>>> headers
+    ) {
+        final Resource resource;
+        String cachedEtag = this.jsonStorage.getEtag(uri);
```
amihaiemil, line 145: @criske just ``etag`` is enough, I think

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
```diff
@@ -0,0 +1,241 @@
...
+        return resource;
+    }
+
+    /**
+     * Store the Resource in json storage only if Etag header is present
+     * and resource's status is HttpURLConnection.OK.
+     * @param uri URI.
+     * @param resource Resource.
+     */
+    private void tryStoreInCache(final URI uri, final Resource resource) {
```
amihaiemil, line 194: @criske just ``storeInCache``. Generally, try to avoid the word ``try`` in method names.

self-core-impl/src/main/java/com/selfxdsd/core/JsonResources.java
```diff
@@ -420,5 +439,24 @@ public JsonArray asJsonArray() {
...
+        }
+
+        @Override
+        public Resource newInstance(
+            final int status,
+            final JsonValue body,
+            final Map<String, List<String>> headers
+        ) {
+            return new JsonResponse(status, body.toString(), headers);
+        }
```
amihaiemil, line 455: @criske Why is this method needed? :D

amihaiemil (changes requested):
@criske looks good, but I have a bunch of comments :D

amihaiemil:
@criske also, will the InMemory json storage work well? I mean, the JSONs from Github/GitLab are pretty big. I imagine with 100 cached requests we will aready fill a lot of memory. One JSON from Github/GitLab must have a few KB, I think :D

self-core-impl/src/main/java/com/selfxdsd/core/CachingJsonResources.java
criske: @amihaiemil For current JsonStorage implementation, no... but if other implementations will keep etag and body in separate keys and support key eviction, we need to double check for a missing cache key.

self-core-impl/src/main/java/com/selfxdsd/core/JsonResources.java
criske: @amihaiemil In CachingJsonResources, the NOT_MODIFIED status Resource response is empty. So we need to "modify" that Resource with an OK status and the cached body, while keeping its headers. Since Resource is an interface, a factory is needed.

```java
 resource.newInstance(
            HttpURLConnection.HTTP_OK,
            Json.createReader(new StringReader(body)).readObject(),
            resource.headers()
        );
```
But now that I'm think of it, a builder based on "this" resource will be more intuitive. The above code will look much better. 
```java
resource.newBuilder().status(HTTP_OK).body(cachedBody).build()
```
Plus I think there is a bug there too: it assumes that `body` is an object, but also it could be an array. It should be
` Json.createReader(new StringReader(body)).read()`.

criske:
> @criske also, will the InMemory json storage work well? I mean, the JSONs from Github/GitLab are pretty big. I imagine with 100 cached requests we will aready fill a lot of memory. One JSON from Github/GitLab must have a few KB, I think :D

@amihaiemil The InMemory should be a temporary solution. We could though implement a LRUCache based on LinkedHashMap and evict the oldest when cache memory is larger than a MB limit. A `Value` entry size in bytes we coud aproximate it based on body length (2 * body.length()). 

But the nice part about this aproach is scalability. We could create a microservice for "json-storage" with three endpoints.
- GET  `/etag?key={base64(uri)}`
- GET `/body?key={base64(uri)}`
- PUT `/store` having [uri, etag, body] as payload

There, the caching impl could be anything: redis, ehcache, even the self-db having a table with "uri" as pk and etag and body as fields.
Now only `self-core` will know about `self-json-storage`, `self-web` or `self-pm` will know nothing about it, not even about caching at all.
The core `JsonStorage` proxy impl should be straightforward:
```java
public String getEtag(final URI uri) {
  Resource res = this.resources.get("{service_uri}/etag?key={base64(uri)}")
  return (res.status() == 200) ? res.asJsonObject().getString("etag") : null 
}
//...
```

self-core-impl/src/main/java/com/selfxdsd/core/JsonResources.java
amihaiemil: @criske newInstance is a fine name, leave it like that :D 

I hade words such as builder, service, instance etc. But newInstance is the least bad haha

amihaiemil:
@criske I think a microservice for this caching is a bit of over-engineering. I think the easiest way is to have a table in our DB with these JSON resources. Can you write a puzzle for implementing this? 

This table should be accessed like all the others, starting from the main Storage interface :D

amihaiemil:
@criske 

> Plus I think there is a bug there too: it assumes that body is an object, but also it could be an array. 

For this leave a puzzle as well, it's good enough for this PR. I want to merge and use this approach, it's quite elegant :D 

BTW, this mechanism should work for GitLab as well? Or at least it will work as before, if GitLab does not support ETag like Github does?

criske:
> @criske
> 
> > Plus I think there is a bug there too: it assumes that body is an object, but also it could be an array.
> 
> For this leave a puzzle as well, it's good enough for this PR. I want to merge and use this approach, it's quite elegant :D
> 
> BTW, this mechanism should work for GitLab as well? Or at least it will work as before, if GitLab does not support ETag like Github does?

@amihaiemil I've already fixed it, via a builder that takes a JsonStructure or string as body. :D. 
All providers supports conditional requests, though not all get requests are conditional..

self-core-impl/src/main/java/com/selfxdsd/core/JsonResources.java
criske: @amihaiemil sorry, I've seen this after my commit ... let me know what to revert/rename  etc :D

amihaiemil:
@criske nice, thanks!

PR #1154, opened 2021-05-14
#1153 Added method Repo#isStarred() via Stars#isStarred().

criske:
pr for #1153 

left puzzle for `GitlabStars#isStarred()` implementation.

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/Stars.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Stars.java
@@ -13,4 +13,11 @@ public interface Stars {
      *  was successful or not.
      */
     boolean add();
+
+    /**
+     * Checks if a repository is starred by the authenticated user.
+     * @return True when starred or false if not or something went wrong
+     * (http error like unauthenticated user etc...).
+     */
+    boolean isStarred();
 }
--- a/self-api/src/main/java/com/selfxdsd/api/Stars.java
+++ b/self-api/src/main/java/com/selfxdsd/api/Stars.java
@@ -15,9 +15,9 @@ public interface Stars {
     boolean add();
 
     /**
-     * Checks if a repository is starred by the authenticated user.
+     * Check if the authenticated user added the star.
      * @return True when starred or false if not or something went wrong
      * (http error like unauthenticated user etc...).
      */
-    boolean isStarred();
+    boolean added();
 }
--- a/self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
@@ -42,9 +42,14 @@ final class GithubStars implements Stars {
     private final Storage storage;
 
     /**
-     * Repo owner holder.
+     * Repo name.
      */
-    private final RepoOwnerHolder holder;
+    private final String repo;
+
+    /**
+     * Owner name.
+     */
+    private final String owner;
 
     /**
      * Ctor.
@@ -61,13 +66,15 @@ final class GithubStars implements Stars {
         this.resources = resources;
         this.starsUri = starsUri;
         this.storage = storage;
-        this.holder = new RepoOwnerHolder(starsUri);
+        List<String> splits = Arrays.asList(starsUri.getPath().split("/"));
+        this.owner = splits.get(splits.size() - 1);
+        this.repo = splits.get(splits.size() - 2);
     }
 
     @Override
     public boolean add() {
         LOG.debug(
-            "Starring Github repository " + this.holder.getOwner()
+            "Starring Github repository " + this.owner
         );
 
         final boolean starred;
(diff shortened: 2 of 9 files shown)


zoeself:

zoeself:

self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
```diff
@@ -78,12 +81,99 @@ public boolean add() {
...
+
+    @Override
+    public String toString() {
+        return this.starsUri.toString();
+    }
+
+    /**
+     * Repo owner holder.
+     */
+    private static class RepoOwnerHolder {
```
amihaiemil, line 140: @criske please remove this class, it's just a utility. Utility classes, helpers, holders, commons etc are what's wrong with today's OOP :) 

I prefer a little duplicated code (that split[] stuff) or maybe a small private utility method.

self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
```diff
@@ -78,12 +81,99 @@ public boolean add() {
...
+        );
+        if (response.statusCode() == HttpURLConnection.HTTP_NO_CONTENT) {
+            isStarred = true;
+            LOG.debug("Repo is starred by current authenticated user.");
+        } else {
+            isStarred = false;
+            if (response.statusCode() == HttpURLConnection.HTTP_NOT_FOUND) {
+                LOG.debug("Repo is not starred by current authenticated user.");
+            } else {
+                LOG.error(
```
amihaiemil, line 119: @criske just a ``warn`` level here pls

self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
```diff
@@ -78,12 +81,99 @@ public boolean add() {
...
+            "Check if Github repository " + this.holder.getRepo()
+                + " is starred by current authenticated user."
+        );
+
+        final boolean isStarred;
+        final Resource response = this.resources.get(
+            URI.create(this.starsUri.toString()),
+            () -> Map.of("Accept", List.of("application/vnd.github.v3+json"))
+        );
+        if (response.statusCode() == HttpURLConnection.HTTP_NO_CONTENT) {
```
amihaiemil, line 111: @criske according to the docs: https://docs.github.com/en/rest/reference/activity#check-if-a-repository-is-starred-by-the-authenticated-user

This endpoint can also return 304 NOT MODIFIED. This means, the request will go through our ConditionalJsonResources mechanism and it will come out as 200 OK and it will break this logic. 

Can you also add ``|| statusCode == OK``? :D

self-api/src/main/java/com/selfxdsd/api/Stars.java
```diff
@@ -13,4 +13,11 @@
      *  was successful or not.
      */
     boolean add();
+
+    /**
+     * Checks if a repository is starred by the authenticated user.
+     * @return True when starred or false if not or something went wrong
+     * (http error like unauthenticated user etc...).
+     */
+    boolean isStarred();
```
amihaiemil, line 22: @criske Please name this method ``added()`` (e.g. Star was added or not).

self-api/src/main/java/com/selfxdsd/api/Stars.java
```diff
@@ -13,4 +13,11 @@
      *  was successful or not.
      */
     boolean add();
+
+    /**
+     * Checks if a repository is starred by the authenticated user.
+     * @return True when starred or false if not or something went wrong
+     * (http error like unauthenticated user etc...).
```
amihaiemil, line 20: @criske "Check if the authenticated user added the star". No need to mention the repo, it's already clear that these Stars belong to a repo.

amihaiemil (changes requested):
@criske see comments :D

self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
criske: > @criske please remove this class, it's just a utility. Utility classes, helpers, holders, commons etc are what's wrong with today's OOP :)

Care to explain why is bad, what's wrong with `java.util.Collections/Arrays` for example?

Anyway I think breaking that class into two member variables initialized at `GithubStars` constructor level would solve the duplicate code problem.

self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
criske: @amihaiemil You can get a 304 only if your request headers has an ["If-None-Match" : <ETag>] entry.

I think the only way to get a 304 is that a prior response from a post/put returned an etag. But in my tests over the wire, starring a repo ( "put") doesn't have an etag in the response header, so I have no idea why 304 is present in their docs :D

Plus ConditionalJsonResources doesn't handle post/put requests.

self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
amihaiemil: Nothings wrong with collections/arrays. 

But this class doesn't implement anything, it's just a utility. 

A class should always implement something, an interface. In fact, the word ``class`` is wrong and misleading. It should be named ``implementation``

self-core-impl/src/main/java/com/selfxdsd/core/GithubStars.java
amihaiemil: @criske but this is a GET, not a PUT and any get will be wored through ConditionalJsonResources, right?

Well, if this GET does not return an Etag, then it is ok.

amihaiemil:
@rultor merge

Issue #1155, opened 2021-05-15
GitlabStars.java: Provide an implementation for... 


zoeself:
The puzzle ``1627255873`` originating from #1153 has to be resolved:

https://github.com/self-xdsd/self-core/blob/d81811c8f26965597ea70759771ec47e6cf39fa7/self-core-impl/src/main/java/com/selfxdsd/core/GitlabStars.java#L39-L39

"Provide an implementation for GitlabStars#isStarred().".

The puzzle was created by cristianpela at ``2021-05-14 12:02:54 +0300``. 
Estimation is ``60 minutes``.
If you have any technical questions, don't ask me, I won't be able to help. Open new issues instead.

zoeself:
@amihaiemil I couldn't find any assignee for this task. This is either because there are no contributors with role ``DEV`` available or because the project does not have enough funds.

Please, make sure there is at least one available contributor with the required role and the project can afford to pay them.

zoeself:
@criske this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-05-25T09:12:27.683745``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

amihaiemil:
@zoeself remove

zoeself:
> @zoeself remove

@amihaiemil ok, I've removed this task from scope. I'm not managing it anymore.

amihaiemil:
@zoeself add

zoeself:
> @zoeself add

@amihaiemil I've just registered this ticket as a task and will assign it to someone soon. Thanks!

zoeself:
@criske this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-05-25T12:52:25.262418``.

Estimation here is ``180 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

[...]

PR #1156, opened 2021-05-15
#1155 Resource paging to assist Github/Gitlab resources.

criske:
pr for #1155

This will be used to implement `GitlabStars#added()`.

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/ResourcePaging.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/ResourcePaging.java
@@ -37,6 +37,44 @@ import java.util.stream.StreamSupport;
 /**
  * Resource Pagination. Implementations should extract the next page link
  * either from headers or from body, depending on the Provider's API.
+ *
+ * <br/>
+ * Usage for Github and Gitlab resources (for Bitbucket will not work because
+ * next page link is inside json body not in response headers - well, it will
+ * work but will not advance to the next page, so other implementation is
+ * needed).
+ * <br/>
+ * Example to get all repos for a user that span over multiple pages.
+ * <br/>
+ * <pre>
+ *     JsonResources res = ...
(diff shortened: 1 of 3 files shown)


zoeself:

criske:
@amihaiemil Unfortunately there is no straight way the check if current user has starred "this" repo. The only endpoint is to get all starred repo by this user (https://docs.gitlab.com/ee/api/projects.html#list-projects-starred-by-a-user)
and then check if this repos is in that list.

Problem is that this requires pagination even though we pas "per_page =100". So this PR come with a Resource Pagination / Headers based implementation that could be used be implement "added()" and in the future for other cases, either for
Github or Gitlab (they use the same format).

zoeself:

self-core-impl/src/main/java/com/selfxdsd/core/ResourcePaging.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.stream.StreamSupport;
+
+/**
+ * Resource Pagination. Implementations should extract the next page link
+ * either from headers or from body, depending on the Provider's API.
+ * @author criske
+ * @version $Id$
+ * @since 0.0.84
+ */
+interface ResourcePaging extends Iterable<Resource> {
```
amihaiemil, line 44: @criske looks interesting, but it's not clear to me how this should be used. Can you extend the JavaDoc with usage examples? :D

amihaiemil (commented):
@criske question

amihaiemil:
@criske very elegant, I like it! Thanks! :D

self-core-impl/src/main/java/com/selfxdsd/core/ResourcePaging.java
criske: @amihaiemil basically it implements this https://docs.github.com/en/rest/guides/traversing-with-pagination (same format is used by Gitlab too)

PR #1176, opened 2021-06-06
ConditionalResource for JsonStorage.

criske:
PR for #1177 

This PR address the problem of multiple db trips when ConditionalJsonResources
checks for etag and body in JsonStorage. By providing a higher abstraction ConditionalResource, an extension of Resource
and therefore changing the original  JsonStorage API, all checks are done at once.

The only downside is that Resource API has to be moved to self-api module in order to be extended by ConditionalResource.

### DIFF
--- a/self-api/src/main/java/com/selfxdsd/api/CachedResource.java
+++ b/self-api/src/main/java/com/selfxdsd/api/CachedResource.java
@@ -30,14 +30,15 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * Conditional {@link Resource}.
- * A Resource is considered conditional if it has a eTag header in its
+ * Cached {@link Resource}.
+ * A Resource is considered cached if it has a eTag header in its
  * response headers.
  * @author criske
  * @version $Id$
  * @since 0.0.8
+ * @see com.selfxdsd.api.storage.JsonStorage
  */
-public interface ConditionalResource extends Resource {
+public interface CachedResource extends Resource {
 
     /**
      * Uri of this resource. This should be also considered the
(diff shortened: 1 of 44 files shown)


zoeself:

zoeself:

criske:
@amihaiemil if this PR ever gets merged, no bounty needed.

self-core-impl/src/test/java/com/selfxdsd/core/ConditionalJsonResourcesTestCase.java
```diff
@@ -232,62 +251,24 @@ public void shouldGetFromRemoteIfRemoteHasChanged(){
...
-        
-        Mockito.verify(storage).store(uri, "etag-124", newBody.toString());
-    }
 
-    /**
-     * Should get Resource from remote if etag is present but 
-     * resource is missing.
-     */
-    @Test
-    public void shouldGetFromRemoteIfResourceEntryIsMissing(){
```
criske, line 244: @amihaiemil This test is not needed anymore. since we get the conditional resource in one go.

amihaiemil:
43 files, record :)))

amihaiemil:
@criske please open an Issue and describe the problem in detail. Then let's break the solution into smaller steps.

I will hate myself if I merge this PR xD

criske:
> 43 files, record :)))

Had no choice here, since ConditionalResource, part of JsonStorage API, extends Resource. 
But Resource is visible only in self-impl. so it had to be moved to self-api. A bunch of files had to be updated though, luckly no logic is getting broken, only the imports are updated.

I'll open an issue.

self-api/src/main/java/com/selfxdsd/api/ConditionalResource.java
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Conditional {@link Resource}.
+ * A Resource is considered conditional if it has a eTag header in its
+ * response headers.
+ * @author criske
+ * @version $Id$
+ * @since 0.0.8
+ */
+public interface ConditionalResource extends Resource {
```
amihaiemil, line 40: @criske can we name it CachedResource? That's what it is actually, a cached resource :D

self-api/src/main/java/com/selfxdsd/api/ConditionalResource.java
```diff
@@ -0,0 +1,213 @@
...
+     * Factory method that creates a ConditionalResource from a regular
+     * Resource.
+     * If the regular can't be conditional
+     * (no ETag header present in its response headers) it returns null.
+     *
+     * @param uri URI associated with the resource.
+     * @param resource Original Resource.
+     * @return ConditionalResource or null.
+     */
+    static ConditionalResource fromResource(
```
amihaiemil, line 74: @criske Please avoid static factory methods. Simply use FromHeaders instead, it should have a constructor which takes Resource

self-core-impl/src/main/java/com/selfxdsd/core/ConditionalJsonResources.java
```diff
@@ -142,48 +143,51 @@ private Resource conditionalGet(
...
                 LOG.debug(
                     "Remote resource body for {} was modified or "
                         + " has an unexpected status code.",
                     uri
                 );
-                resource = remoteResource;
-                this.storeInCache(uri, resource);
+                final ConditionalResource newConditional = ConditionalResource
+                    .fromResource(uri, remoteResource);
+                if (newConditional != null) {
```
amihaiemil, line 167: @criske here we will just have ``if(cached.eTag() != null)``

self-core-impl/src/main/java/com/selfxdsd/core/ConditionalJsonResources.java
```diff
@@ -142,48 +143,51 @@ private Resource conditionalGet(
...
+                resource = stored;
             } else {
                 LOG.debug(
                     "Remote resource body for {} was modified or "
                         + " has an unexpected status code.",
                     uri
                 );
-                resource = remoteResource;
-                this.storeInCache(uri, resource);
+                final ConditionalResource newConditional = ConditionalResource
```
amihaiemil, line 165: @criske I think this variable can be renamed to ``cached``

self-api/src/main/java/com/selfxdsd/api/ConditionalResource.java
```diff
@@ -0,0 +1,213 @@
...
+    private static String findEtag(final Resource resource) {
+        return resource
+            .headers()
+            .entrySet()
+            .stream()
+            .filter(entry -> entry.getKey()
+                .equalsIgnoreCase("ETag"))
+            .flatMap(entry -> entry.getValue().stream())
+            .findFirst()
+            .orElse(null);
```
amihaiemil, line 107: @criske this should be inside FromHeaders.eTag()

amihaiemil (changes requested):
@criske a few comments :D

self-api/src/main/java/com/selfxdsd/api/ConditionalResource.java
criske: @amihaiemil If we're doing this then, when Resource is not cacheable (no etag in headers) we need to throw an exception in constructor instead and therefore making things a little bit ugly in ConditionalResources (we need to catch it). This is why FromResource constructor is private and let this factory to take control and return null otherwise. Also the factory will not create a new CachedResource if the provided Resource is already cached.

However I think it will be cleaner and more intuitive if we're moving the factory to `FromResource` class scope or, even better, moving the whole class FromResource along with its factory to self-impl. I don't think we'll ever use this class outside of self-impl (in self-storage surely not).

self-core-impl/src/main/java/com/selfxdsd/core/ConditionalJsonResources.java
criske: @amihaiemil Doesn't make sense to have CachedResource `eTag` null, that's why is cached in the first place :D

self-api/src/main/java/com/selfxdsd/api/ConditionalResource.java
criske: @amihaiemil I agree, this could be inlined... but etag is already provided with CachedResource constructor. Again, the factory handles this.

self-api/src/main/java/com/selfxdsd/api/ConditionalResource.java
amihaiemil: @criske ok, then let's leave this as it is :D

self-api/src/main/java/com/selfxdsd/api/CachedResource.java
```diff
@@ -55,65 +56,60 @@
...
-        if (resource instanceof ConditionalResource) {
-            conditional = (ConditionalResource) resource;
+        final CachedResource cached;
+        if (resource instanceof CachedResource) {
+            final CachedResource candidate = (CachedResource) resource;
+            if (!candidate.uri().equals(uri)) {
+                cached = new FromResource(candidate, candidate.etag(), uri);
+            } else {
+                cached = candidate;
+            }
```
criske, line 86: @amihaiemil Here a small bug is fixed. Now it makes sure to have the same uri too in order to return the  same instance,
 otherwise it mutates the resource with the new uri. 
Most likely this corner case we'will never encounter in prod...  :D

PR #1197, opened 2021-08-04
#1996 started on Project renaming and left puzzles

### LINKED ISSUE #1196: Start implementing Project Renaming
Users may rename their Github (maybe also GitLab) repos, so we must catch this webhook event and also rename the Project on our side.

amihaiemil:
Fixes #1196


zoeself:

zoeself:

Issue #1237, opened 2021-10-08
Tasks.java: Implement {@link Tasks#update... 


zoeself:
The puzzle ``-1257137381`` originating from #1233 has to be resolved:

https://github.com/self-xdsd/self-core/blob/9edf5cbcd81d25d04498af7042305f458b4dde1d/self-api/src/main/java/com/selfxdsd/api/Tasks.java#L30-L31

"Implement {@link Tasks#updateEstimation(Task, int)} for ContractTasks, ContributorTasks, and ProjectTasks.".

The puzzle was created by cristianpela at ``2021-10-06 10:58:05 +0300``. 
Estimation is ``30 minutes``.
If you have any technical questions, don't ask me, I won't be able to help. Open new issues instead.

PR #1241, opened 2021-10-12
#1237 Task update estimation for Contract/Contributor/ProjectTasks.

criske:
PR for #1237


zoeself:

PR #1250, opened 2021-10-15
#1245 StoredProjectManager#issueLabelsChanged check if task exists.

### LINKED ISSUE #1245: StoredProjectManater.issueLabelsChanged(...) Check If Task Exists
We should check whether the tasks exists (!= null) [here](https://github.com/self-xdsd/self-core/blob/master/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java#L656).

When opening a new Issue which has a label, Github might sent us the ``labeled`` event webhook **before** the ``opened`` event (and in this case, the Task doesn't exist on our side, yet).

criske:
Fixes #1245


zoeself:

amihaiemil:
@criske thanks, looks good!

Issue #1271, opened 2021-10-30
StoredProjectManagerTestCase.java: Unignore/update this test onc... 


zoeself:
The puzzle ``904709704`` originating from #1265 has to be resolved:

https://github.com/self-xdsd/self-core/blob/3edc2727a3286993bd2eb67186f8ad62ae55a894/self-core-impl/src/test/java/com/selfxdsd/core/managers/StoredProjectManagerTestCase.java#L1832-L1833

"Unignore/update this test once the whole functionality is refactored to use Steps.".

The puzzle was created by amihaiemil at ``2021-10-30 10:34:14 +0300``. 
Estimation is ``60 minutes``.
If you have any technical questions, don't ask me, I won't be able to help. Open new issues instead.

zoeself:
@amihaiemil this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-11-09T07:40:12.077754``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

amihaiemil:
@zoeself resign

zoeself:
> @zoeself resign

@amihaiemil ok, I'll find someone else to take care of it.

zoeself:
@criske this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-11-13T09:31:44.931423``.

Estimation here is ``60 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

zoeself:
@criske Don't forget to close this ticket before the deadline (2021-11-13T09:31:45). You are past the first half of the allowed period.

amihaiemil:
@zoeself remove

zoeself:
> @zoeself remove

[...]

Issue #1269, opened 2021-10-30
StoredProjectManager.java: Send a comment after updating... 


zoeself:
The puzzle ``707400492`` originating from #1265 has to be resolved:

https://github.com/self-xdsd/self-core/blob/3edc2727a3286993bd2eb67186f8ad62ae55a894/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java#L55-L56

"Send a comment after updating a task's estimation inside issueLabelsChanged.".

The puzzle was created by amihaiemil at ``2021-10-30 10:34:14 +0300``. 
Estimation is ``30 minutes``.
If you have any technical questions, don't ask me, I won't be able to help. Open new issues instead.

zoeself:
@amihaiemil this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-11-09T07:40:05.497638``.

Estimation here is ``30 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

amihaiemil:
@zoeself resign

zoeself:
> @zoeself resign

@amihaiemil ok, I'll find someone else to take care of it.

zoeself:
@criske this is your task now, please go ahead. Deadline (when this ticket should be closed) is ``2021-11-13T09:31:41.852543``.

Estimation here is ``30 minutes``, that's how much you will be paid.

Remember, you don't have to solve everything in this ticket - you can solve it [partially](https://docs.self-xdsd.com/microtasking.html) and leave ``todo`` markers in the code, which will become future tasks.

If you have any questions don't ask me, I'm not a technical person. Open new tickets instead.

If you don't have time or simply don't want to work on this, you can always [resign](https://docs.self-xdsd.com/projectmanager.html#commands).

zoeself:
@criske Don't forget to close this ticket before the deadline (2021-11-13T09:31:42). You are past the first half of the allowed period.

criske:
@zoeself waiting for #1268

zoeself:
> @zoeself waiting for #1268 

[...]

PR #1283, opened 2021-11-10
#1269 Send a comment after updating task estimation.

criske:
This logic will be tested along with #1271.

Pr for #1269

### DIFF
--- a/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
+++ b/self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
@@ -49,8 +49,6 @@ import java.util.function.Supplier;
  * @checkstyle ClassFanOutComplexity (1000 lines)
  * @checkstyle ClassDataAbstractionCoupling (1000 lines)
  * @since 0.0.1
- * @todo #1265:30min Send a comment after updating a task's estimation inside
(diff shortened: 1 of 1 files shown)


zoeself:

 If this PR solves a ``todo`` from the code, please don't forget to remove it.

criske:
@amihaiemil no bounty for this :D

self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
```diff
@@ -665,8 +663,7 @@ public void issueLabelsChanged(final Event event) {
             isNotClosed -> new TaskIsRegistered(
                 new IssueEstimationChanged(
                     new UpdateTaskEstimation(
-                        sendReply -> LOG.debug("Send updated estimation"
```
criske, line 668: @amihaiemil btw, is there a "Log" step? It will be nice to retain this info too.

zoeself:

self-core-impl/src/main/java/com/selfxdsd/core/managers/StoredProjectManager.java
amihaiemil: @criske nope, but I'm pretty sure SendReply does some logging internally :D

